PR #5, opened 2015-10-19
Call one SOAP endpoint for all requests save end-user username/password validation

brandocalrissian:
We use the RegisterSignService endpoint for all service calls other than
authenticating the end user by username/password. Even the "system" user
authentication is with RegisterSignService.

Change the default endpoint URI to RegisterSignService and override the
user authentication endpoint to RegisterAuthService.


baccigalupi:
Thanks for catching this!

PR #13, opened 2016-01-05
First draft of JSON Schema

scottdchristian:
The users have requested defined parameters for the data they will be
submitting.  This document has the goal of providing such guidance.  I
placed the schemas in a folder in the db folder.  However it should go
in ""https://e-manifest.18f.gov/schemas/manifestSchema20160105.json"

### DIFF
--- a/db/schemas/manifestSchema20160105.json
+++ b/db/schemas/manifestSchema20160105.json
@@ -1,0 +1,367 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "id":"https://e-manifest.18f.gov/schemas/manifestSchema20160105.json",
+    "title": "e-Manifest JSON Schema",
+    "description": "First draft of JSON Schema.  I fully expect this to be editted and revised.  Resource used to make this was http://spacetelescope.github.io/understanding-json-schema/index.html",
+
+    "definitions":{
+        "address": {
+             "type": "object",
+            "properties":{
+                "address_1":   {
+                    "description": "Address line 1",
+                    "type": "string",
+                    "maxLength":30 ,
+                    "minLength":0 
+                },
+                "address_2":   {
+                    "description": "Address line 2",
+                    "type": "string",
+                    "maxLength":30 ,
+                    "minLength":0 
+                },
+                "city":   {
+                    "description": "City name",
+                    "type": "string",
+                    "maxLength":25 ,
+                    "minLength":0 
+                },
+                "state":   {
+                    "description": "Two character abbreviation for State or Province",
+                    "type": "string",
+                    "maxLength":2,
+                    "minLength":2, 
+                    "COMMENT":" add Enummerated value"
+                },
+                "zip_code":   {
+                    "description": "Postal Code",
+                    "type": "string",
+                    "maxLength":14,
+                    "minLength":0, 
+                    "COMMENT":" add Regular Expression"
+                }
+            }
+        },
+
+        "TelephoneNumberDataType":  {
+            "description": "15 alphanumberic string accounting for international numbers e.g. +001 1234321234, to be honest this is just a safety feature, most will be standard 10 digit",
+            "type": "string",
+            "maxLength":15,
+            "minLength":0, 
+            "COMMENT": "add Regular Expressions"
+        },
+        "signature_info":{
+             "type": "object",
+            "properties":{
+                "name": {
+                    "description": "Name of the signatory",
+                    "type": "string",
+                    "maxLength": 30,
+                    "minLength": 0
+                },
+                "month": {
+                    "description": "Two digit month of signature",
+                    "type": "integer",
+                    "maximum": 2,
+                    "minimum": 2
+                },
+                "day": {
+                    "description": "Two digit day of signature",
+                    "type": "integer",
+                    "maximum": 2,
+                    "minimum": 2
+                },
+                "year": {
+                    "description": "Four digit year of signature",
+                    "type": "integer",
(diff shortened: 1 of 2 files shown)


jessieay:
why is there a date on this schema? seems like we could just call it `manifestSchema` and update its content as required?

scottdchristian:
Thanks, actually force of bad habit.

A major point of GitHub is version control.  You are right I should have just named it how we would use it rather than my desktop version control.  I will rename and edit in the morning then refresh.

Thanks again,
Scott

db/schemas/manifestSchema20160105.json
```diff
@@ -0,0 +1,367 @@
...
+                    "description": "Postal Code",
+                    "type": "string",
+                    "maxLength":14,
+                    "minLength":0, 
+                    "COMMENT":" add Regular Expression"
+                }
+            }
+        },
+
+        "TelephoneNumberDataType":  {
```
jeremiak: Looks like we're using snake case everywhere else so this field should be `telephone_number_data_type`

jessieay: I've always used snake case when creating JSON APIs with Ruby because that is a Ruby convention. 

I recently read, however, that it is preferred to use snake case since JSON is JavaScript and js convention is to use CamelCase. 

I am fine with either as long as we are consistent.

db/schemas/manifestSchema20160105.json
```diff
@@ -0,0 +1,367 @@
...
+
+    "definitions":{
+        "address": {
+             "type": "object",
+            "properties":{
+                "address_1":   {
+                    "description": "Address line 1",
+                    "type": "string",
+                    "maxLength":30 ,
+                    "minLength":0
```
jeremiak: Should these actually be something like `"minLength": 1` so that we get a validation error if these values are not supplied?

db/schemas/manifestSchema20160105.json
pkarman: +1 for consistency.

It's true that Google and others consistently use camelCase for JSON. schema.org does that as well.

More tellingly, since this schema purports to adhere to json-schema.org conventions, the examples there all camelCase.

pkarman: having said all that, I don't think it really matters to consumers of the API, as long as it is consistent and well-documented.

scottdchristian: I will go with consistency.  For this project the initial data was written in snake_case, since users have already started integrating the JSON and I do not see a compelling reason to change the field names now.  

With that said, my next iteration (after this one) is to add new fields to the Schema.

db/schemas/manifestschema.json
```diff
@@ -0,0 +1,356 @@
...
+                "zip_code":   {
+                    "description": "Postal Code",
+                    "type": "string",
+                    "maxLength":14, 
+                    "COMMENT":" add Regular Expression"
+                }
+            }
+        },
+
+        "telephonenumberdatatype":  {
```
jessieay, line 42: should be `telephone_number_data_type`

db/schemas/manifestschema.json
```diff
@@ -0,0 +1,356 @@
...
+                        "description": "(18c) Alternate facility (or Generator) signature information",
+                        "$ref": "#/definitions/signature_info" 
+                    },
+                        "address": {
+                        "description": "(18b) Alternate facility address as presented in RCRAInfo",
+                        "$ref": "#/definitions/address" 
+                    }
+                    }
+                  },
+
```
jessieay, line 270: let's remove these extra newlines

db/schemas/manifestschema.json
```diff
@@ -0,0 +1,356 @@
...
+
+                "manifest_items": {
+                    "type": "array",
+                    "items": [{
```
jessieay, line 274: I think the opening curly brace on this line (and its closing curly brace) are not needed since this is an array. Or is it array of objects and there is one object?

pkarman, line 274: I would recommend we just run the whole file through a JSON pretty printer so there's no question of manual formatting (and makes diffs better later on).

jessieay, line 274: :+1:

db/schemas/manifestschema.json
scottdchristian, line 42: OK, but we need a plan to make this change.  The name included is the name of the field in industry.html and the one posted in our sample.  Let's discuss at story time.  FWIW, I am not opposed to changing the names much.  We'll be adding new fields anyway.

db/schemas/manifestschema.json
scottdchristian, line 270: Newbie question, which lines to be specific?

jessieay, line 270: 270 and 271 :)

scottdchristian, line 270: Thanks

db/schemas/manifestschema.json
scottdchristian, line 274: Looking at the JSON and form it is an array of one object that get's repeated.  manifest_items is a section of the form and all those pieces are part of that section.  A manifest can have one manifest item or many manifest items. On the form items 9a-13 (http://www3.epa.gov/epawaste/hazard/transportation/manifest/pdf/newform.pdf)

scottdchristian, line 274: I didn't answer the question directly:  Yes its an array of one object

jessieay:
Going to merge as-is so I can start working on validating that our JSON matches this schema

PR #14, opened 2016-01-05
Elasticsearch

pkarman:
This PR adds several features:
- search with Elasticsearch
- background worker with Sidekiq
- search API requires `q` (query) or `aq` (advanced query) params
- rspec uses DatabaseCleaner
- `rake serve` now uses foreman to run web and worker together
- the `/sidekiq` web endpoint requires configurable basic authentication

See stories: 
- https://trello.com/c/xtxYU5Mp
- https://trello.com/c/rlERjxud


spec/search_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it "searched by structured query" do
+      populate_manifests
+      search_for_advanced({ "content.generator.name" => "generator" })
+      expect(JSON.parse(last_response.body)["total"]).to eq 10
+    end
+  end
+
+  def populate_manifests
+    populator = Populator.new(Manifest)
+    populator.run(10)
```
jessieay: what do you think abut doing 2 instead? Seems like that we don't need 10 records for this spec to be helpful / effective

spec/query_dsl_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+          query: "(foo) AND (color:(green))",
+          default_operator: "and"
+        },
+      },
+      size: QueryDSL::MAX_RESULTS,
+      from: 0
+    })
+  end
+
+  # TODO pass with current_user and test filter
```
jessieay, line 16: what does this mean? should we create a card for it so we can have more details?

spec/fielded_search_spec.rb
```diff
@@ -0,0 +1,47 @@
+describe FieldedSearch do
+  it "#to_s" do
+    fs = FieldedSearch.new({
+      foo: "bar"
+    })
+    expect(fs.to_s).to eq "foo:(bar)"
+  end
```
jessieay, line 7: I think newlines between these blocks would make this spec file easier to read

spec/search_spec.rb
pkarman: I anticipate needing to write tests for sorting and ranking, so having more than 2 by default would be useful.

jessieay: we could do more than 10 for those -- perhaps this method could take a number as an optional argument and then create the number of records passed in?

spec/query_dsl_spec.rb
pkarman, line 16: this card sums it up: https://trello.com/c/0RRTt2mN

right now there is no sense of a `current_user` in the app, but when there is, we'll want to be able to add a ES filter to limit search results by authorization.

Gemfile
```diff
@@ -15,7 +18,15 @@ gem 'json-patch'
 
 gem 'savon-multipart'
 
+gem 'sidekiq'
+gem "kaminari"
```
jessieay: seems like there's a combo of single and double quotes in here -- any sense on if one or the other is used more in this project? if so, let's move toward whatever that is

README.md
```diff
@@ -32,6 +34,8 @@ Don't edit the `public/` folder directly! Jekyll will overwrite everything when
 - To build the static site, run `rake build`.
 - To build and serve the app, run `rake serve`.
 - To deploy to 18F's cloud, run `rake deploy`.
+- To add dummy data for developing against, run `rake populate:manifests`.
+- To (re)build the Elasticsearch index, run `rake search:index`.
```
jessieay: so many steps! just created a card for automating all of this by putting it in a script https://trello.com/c/RieEsKeK

_static/assets/js/app.js
```diff
@@ -184,95 +184,115 @@
...
-        {
-           items.push(item);
-        }
+      if ($scope.data.tsdf_name) {
+        adv_query['content.designated_facility.name'] = $scope.data.tsdf_name;
       }
-
-      $scope.filtered = items;
-    };
+      // TODO state/province, date
```
jessieay, line 265: card for this?

spec/query_dsl_spec.rb
jessieay, line 16: ah ok, so perhaps this code comment would be more helpful as a comment in that Trello card so that whoever implements `current_user` knows to also update this file

lib/query_dsl.rb
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  def add_filter
+    searchdsl = self
+    if apply_authz?
+      @dsl.filter = Filter.new
+      @dsl.filter do
+        bool do
+          must do
+            term "subscribers.id" => searchdsl.current_user.id.to_s
```
jessieay, line 102: since `current_user` can be anything but `nil` to get to this point  (`current_user != nil`), isn't there a risk that we would get a `noMethodErrror` when calling `id` ?

spec/spec_helper.rb
```diff
@@ -45,6 +91,22 @@
     mocks.verify_partial_doubles = true
   end
 
+  config.before(:suite) do
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.clean_with(:truncation)
+    start_es_server unless ENV['ES_SKIP']
```
jessieay, line 71: what about doing this before specific specs rather than before the whole test suite? (wondering if there will be problems trying to run individual specs that require the es_server if we only call this here)

config.ru
```diff
@@ -2,10 +2,15 @@
...
+require 'sidekiq/web'
 
 use Rack::TryStatic,
   :root => 'public',
   :urls => %w[/],
   :try => ['.html', 'index.html', '/index.html']
 
-run App
+Sidekiq::Web.use Rack::Auth::Basic do |username, password|
+  username == ENV["SIDEKIQ_USERNAME"] && password == ENV["SIDEKIQ_PASSWORD"]
```
jessieay, line 13: add these to `.env.example` ?

config/elasticsearch.rb
```diff
@@ -0,0 +1,46 @@
...
+if ENV["RACK_ENV"] == "production"
+  vcap = ENV["VCAP_SERVICES"]
+  vcap_config = JSON.parse(vcap)
+  vcap_config.keys.each do |vcap_key|
+    if vcap_key.match(/elasticsearch/)
+      es_config = vcap_config[vcap_key]
+      es_client_args[:url] = es_config[0]["credentials"]["uri"]
+    end 
+  end 
+elsif ENV["RACK_ENV"] == "test"
```
jessieay, line 24: should we define this in `.env.test` ?

lib/app.rb
```diff
@@ -44,7 +37,11 @@ class App < Sinatra::Base
   # Search for Manifests
   get '/api/:version/manifest/search' do
     content_type :json
-    Manifest.all.to_json
+    if !params[:q] && !params[:aq]
```
jessieay, line 40: what about giving this param a more verbose name like `query`? (not sure what `aq` means, but a longer name might help with that as well)

lib/chores/populator.rb
```diff
@@ -0,0 +1,177 @@
...
+
+class Populator
+  def initialize(model, n_records=100)
+    @model = model
+    @n_records = (n_records || 100).to_i
+  end
+
+  def run(n_records=@n_records)
+    method_name = "make_#{@model.to_s.downcase}_record".to_sym
+    unless respond_to?(method_name, true)
```
jessieay, line 11: what happens when you don't have this? seems like it would give you a `noMethodError` without it, but guessing the error msg was unclear somehow?

spec/spec_helper.rb
pkarman, line 71: I often run single spec file tests like this:

```
% ES_SKIP=1 rspec spec/query_dsl_spec.rb
```

and it works as expected (no Elasticsearch test cluster is started/stopped).

config/elasticsearch.rb
pkarman, line 24: it is already defined in `spec/spec_helper.rb`

pkarman, line 24: also, `.env.test` is not loaded till after `RACK_ENV` is evaluated. Chicken and egg to place it there.

lib/fielded_search.rb
```diff
@@ -0,0 +1,63 @@
...
+    to_s.present?
+  end
+
+  def value_for(key)
+    if field_pairs && field_pairs.key?(key)
+      field_pairs[key]
+    end
+  end
+
+  def to_s
```
jessieay, line 19: defining `present?` and `to_h` and `to_s` feels a bit icky to me because those are methods Ruby gives us. Would it make sense to name these methods differently?

lib/app.rb
pkarman, line 40: I was aiming for compatibility with google.com: https://www.google.com/#q=search+query+param+name

lib/models/manifest.rb
```diff
@@ -1,3 +1,56 @@
+require_relative '../query_dsl'
+
 class Manifest < ActiveRecord::Base
+  include Elasticsearch::Model
+
+  ActiveRecord::Base.raise_in_transactional_callbacks = true
+
+  after_commit on: [:create] do
```
jessieay, line 8: thoughts on doing this reindexing at `lib/app.rb` instead of in callbacks? always nice to avoid callbacks...

lib/chores/populator.rb
pkarman, line 11: I think of it like an abstract class. There might be a valid model called `Foo` but no populator support for it (yet). Hence the custom error message.

lib/query_dsl.rb
pkarman, line 102: a great deal of risk, yes, which I am counting on triggering an error once we try to add `current_user` support. I consider that a feature. :)

lib/tasks/search.rake
```diff
@@ -0,0 +1,245 @@
...
+      :klass => klass, 
+      :idx_name => klass.index_name,
+      :nprocs => nprocs.to_i, 
+      :batch_size => batch_size.to_i, 
+      :max => max, 
+      :force => ENV['FORCE']
+    })
+  end
+
+  def create_mindex(opts)
```
jessieay, line 26: this method is doing a lot -- what do you think about putting this code / logic into a testable class? Rake tasks are a pain to test, so don't usually test them, but I do usually like to have the invoke well-tested classes.

lib/app.rb
jessieay, line 40: oh ok - why is that?

spec/spec_helper.rb
jessieay, line 71: my workflow is to run specs automatically from within the code using a vim-rspec plugin. Yoz does something similar with atom. running the above, while possible, would not work with our current rspec workflows. 

the larger concern I have is that running a single spec where you _do_ want to start the es server would not work unless that spec file is run within the suite. does that make sense? perhaps I am wrong but I would expect that specs requiring the es server would only work when run within the test suite, which would present a problem for all spec running workflows.

lib/query_dsl.rb
jessieay, line 102: that makes sense. it does feel a bit confusing to create search filters based on a current user when we don't have that yet, but I agree that errors here would make sense given the current world we are living in

spec/search_spec.rb
pkarman: good idea. implemented in d57c578

spec/fielded_search_spec.rb
pkarman, line 7: as you wish, in bfbe9b2

spec/query_dsl_spec.rb
pkarman, line 16: thanks, added.

spec/spec_helper.rb
pkarman, line 71: Have you tried it? I just read through the docs of the vim-rspec plugin (https://github.com/thoughtbot/vim-rspec) and it seems to do the same thing I am doing manually.

My understanding is that `:suite` is shorthand for "all the individual tests that will run in this process" and so is invoked every time the `rspec` command executes.

config.ru
pkarman, line 13: good call. done in 4dd14c0fb373136fc3a0c5a14d32c7ce15d94ac4

lib/app.rb
pkarman, line 40: search applications historically use the `q` param for the query string. I try to keep with that convention. Here are some reference documents.

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-uri-request.html
https://wiki.apache.org/solr/CommonQueryParameters

My understanding is that short URL parameter names are used because Back In the Day there were limits to URL length and each extra byte in the URL required that much more bandwidth.

lib/fielded_search.rb
pkarman, line 19: `present?` is a method specific to Ruby on Rails, not Ruby.

```
[peterkarman@peterkarman-gsa:~/src/C2]$ irb
2.2.3 :001 > f = nil
 => nil 
2.2.3 :002 > f.present?
NoMethodError: undefined method `present?' for nil:NilClass
    from (irb):2
    from /Users/peterkarman/.rvm/rubies/ruby-2.2.3/bin/irb:15:in `<main>'
```

I chose that name intentionally however, because of how it is used in RoR, to indicate whether the object is not `blank?`.

The same is true of `to_h` and `to_s` -- those are special method names that Ruby programmers know, and they act consistently with this particular object type, which uses a delegation pattern. If you want a `FieldedSearch` object to stringify correctly and automatically, overriding `to_s` is the way to do it.

lib/models/manifest.rb
pkarman, line 8: using callbacks is the preferred way for indexing. See https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model#user-content-asynchronous-callbacks

lib/tasks/search.rake
pkarman, line 26: it's probably worth splitting into its own gem since I have used this same code on other projects. I will take that as a todo.

_static/assets/js/app.js
pkarman, line 265: https://trello.com/c/El5kKAkI

spec/spec_helper.rb
jessieay, line 71: ah that is interesting! I did not try it, I always thought that `suite` was before running all specs rather than each one. Very interesting!

lib/fielded_search.rb
jessieay, line 19: ah yes, my bad. I guess that is part of the reason I find these confusing -- It feels like we are monkeypatching even though we aren't. 

like in `organization.rb` where we defined `self.all` and `self.find` which are Rails methods. That was just a Ruby class, but defining the same methods at Rails but defining them differently proved to be quite confusing when refactoring.  (https://github.com/18F/C2/pull/870/files) 

What do you see as the downside as giving these methods more specific / descriptive names?

lib/models/manifest.rb
jessieay, line 8: ok - then let's just make sure the methods using for this callback are private since we aren't using them outside of this class.

pkarman:
@brandocalrissian this is now in sync with master.

brandocalrissian:
Thanks, @pkarman! Looking at it now.

PR #18, opened 2016-01-06
Add webmock

jessieay:
- Tests should never hit live APIs or other sites
- Webmock ensures that they do not

https://trello.com/c/4bKQPSW8


pkarman:
this effectively turns off all the testing of the CDX service. I agree that it would be good not to hit it regularly as part of the rspec suite. Should we instead have some kind of CDX mock service, so we can actually exercise the full cycle? 

@brandocalrissian thoughts?

jessieay:
Yeah that's a good idea.

This came up today because I wrote some request specs for authentication
requests with a bad password and we got locked out of our account after I
ran the specs three times.

I figure we can add webmock here and then I can add a mock service in the
Branch where I added the auth request specs because I should now get a
webmock error there rather than hitting Cdx and causing problems with our
dev account.

pkarman:
@jessieay merge conflict, alas. once fixed I can merge.

jessieay:
@pkarman rebased!

PR #19, opened 2016-01-06
Add travis config file

jessieay:


jessieay:
Note: we should add the appropriate env vars to cloud foundry before deploying. Can someone add me to CF?

.sample.env
```diff
@@ -0,0 +1,2 @@
+CDX_USERNAME="change_me"
+CDX_PASSWORD="change_me"
```
pkarman: don't we already have a `.env.example`? Why this new file?

README.md
```diff
@@ -10,7 +10,7 @@ Don't edit the `public/` folder directly! Jekyll will overwrite everything when
 
 ## Running Locally
 
-0. Copy the secret.rb.template to secret.rb. See below CROMERR Signing
+0. Copy the `.sample.env` to `.env` See below CROMERR Signing
 0. Install Ruby
```
pkarman: `.env.example` ?

.gitignore
```diff
@@ -11,8 +11,5 @@ cf-ssh.yml
 # Ignore emacs autosave
 **/*~
 
-# Ignore configuration secrets file
-secret.rb
-
 # Ignore local envvars
-.env
```
pkarman, line 18: we don't want any `.env` file checked in do we? Why zap that line?

pkarman:
@jessieay only real comment is about `.sample.env` vs existing `.env.example`. Otherwise looks fine.

jessieay:
@pkarman updated! missed that we had an `.env.example` file already -- moved example env vars there and removed `.sample.env`

.gitignore
jessieay, line 18: the new instructions are to use `.env.development` instead of `.env` locally. The reason for this is that I found that having env vars in a `.env` file as overriding env vars in`.env.test` and `.env.development` -- when I defined `CDX_USERNAME` in `.env`, for example, the env var spec failed. I suspect this is a load order issue but an easy way around it was to use env-specific env files.

pkarman:
@jessieay oops, more conflict.

jessieay:
@pkarman re-updated

PR #20, opened 2016-01-06
refactor API documentation to use swagger.io

pkarman:
Story: https://trello.com/c/Ifvzi0Kj

This refactor creates a new `/api` endpoint and site nav. That endpoint lists Documentation, Examples and Diagnostics.


jessieay:
is there any way we could generate the swagger docs without adding all of this code? I understand if not, but reviewing 32k lines seems like a bit much. I could scroll through and just give the thumbs up, but I'd no be giving an actual review...

pkarman:
@jessieay you can ignore everything in `_static/api-documentation/*` -- that's all part of the swagger-ui code I just downloaded from their site. The only thing in there that isn't 3rd party is `swagger.json`, some customizations to `index.html` and the `css/emanifest.css` file.

jessieay:
@pkarman is there a way to pull in all of that boilerplate without pull the code in? 

Imagining that having all of these files will be problematic for grepping the project. Also, I can imagine future developers diving in and editing these files, which is not ideal because it will make updating swagger difficult if we add customizations directly to the lib code.

If there is a way to pull in swagger boilerplate with a library and then add _just_ the customizations that would be :+1: :+1:

pkarman:
This isn't rails (yet) or we might be able to use https://github.com/d4be4st/swagger-ui_rails

Still, these files are like the US design standard CSS/JS files. They need to be checked in and available to the Jekyll site builder. And in this case, there are some deviations from the standard swagger-ui CSS in order to make the docs integrate with the US design look/feel.

These are just 3rd party assets, no harder to skip in a grep than bootstrap or angular or any other 3rd party files.

jessieay:
OK perhaps we should merge as-is and then use that Rails gem when we move to Rails. Agree that these 3rd party assets are no different from other front end assets, but that's one reason why I don't love using huge libs like bootstrap.

jessieay:
QA instructions via Peter:

there should be a nav link to ‘API'
that includes lines to examples, documentation and diagnostics
the examples is howto-style
the documentation is the swagger stuff

jessieay:
OK, the `method_codes` endpoint is now working. 

`GET /manifest/:manifest_tracking_number` is still not working when I enter in the id "123" (curl http://localhost:9292/api/0.1/manifest/123 _does_ work locally) -- but I figure we can merge this and the next step is to make sure all of the swagger endpoints actually work?

pkarman:
+1 to merge and find bugs after.

PR #26, opened 2016-01-07
Add custom Rspec matcher for Manifest GET requests

jessieay:
- Use `json_schema` gem https://github.com/brandur/json_schema
- This is preferred because it favors format of json schema that
  committee gem expects, which is what we are going to use to validate
  JSON in the app: https://github.com/interagent/committee
- Use ActiveModel Serializers to generate custom json for Manifest
- Validate presence of top-level JSON keys
- Will validate nested keys next


pkarman:
This looks great. I like ManifestSerializer class particularly.

I am going to wait to merge this till #14 gets merged though, because I think there will be some incidental merge conflicts.

jessieay:
@pkarman rebased against master

lib/schemas/get_manifest.json
```diff
@@ -0,0 +1,369 @@
+{
```
pkarman, line 1: This looks like a slightly tweeked copy of `post_manifest.json` -- was there no way to re-use the basic manifest schema w/o copying it over?

jessieay, line 1: combining manifest files is a bit tricky. Trying to follow some of the work here https://github.com/rails-api/active_model_serializers/pull/1270/files and in various SO threads but so far no luck. Thoughts on merging now and fixing later? 

It is something like  "$ref": "file://lib/schemas/manifest_content.json#" but so far I can't get anything to work...

pkarman:
@jessieay I would rather that we figure out a way to reduce duplication of schema definitions. Too hard to keep in sync. This will need to be reworked against master now anyway.

I will try and figure out how to use the JSON `$ref`. If we can't do it, make there's a way to make building the schemas from a template possible.

jessieay:
@pkarman OK, I will close this for now. The PR I liked to in my comment uses ref, but there is something I wasn't able to get working

PR #27, opened 2016-01-08
Change API version to V0, Update routes

jessieay:
This is an incremental API refactor. Trying to make route more standard REST routes.


_static/api-examples.md
```diff
@@ -126,7 +126,7 @@ You must authenticate with the CDX CROMERR service for signing a manifest. Value
 ```bash
 curl -i -X POST -H 'Content-Type: application/json' \
   --data @auth-creds.json \
-  https://e-manifest.18f.gov/api/0.1/user/authenticate
+  https://e-manifest.18f.gov/api/v1/users/authentications
```
pkarman: is plural appropriate here? A request only ever operates on a single `user`.

I realize we want to keep verbs out of the URL itself, but `authentications` feels a little awkward to me. REST semantics aren't gospel.

_static/api-examples.md
```diff
@@ -152,7 +152,7 @@ You can sign a manifest with either the e-Manifest ID:
 ```bash
 curl -i -X POST -H 'Content-Type: application/json' \
   --data @sign-manifest-emanifestid.json \
-  https://e-manifest.18f.gov/api/0.1/manifest/sign
+  https://e-manifest.18f.gov/api/v1/manifests/:manifest_id/sign
```
pkarman: `signature` instead of `sign`? If we're trying to keep the URL verb-free.

_static/api-examples.md
```diff
@@ -166,7 +166,7 @@ You may also sign with the Manifest Tracking Number:
 ```bash
 curl -i -X POST -H 'Content-Type: application/json' \
   --data @sign-manifest-manifesttrackingnumber.json \
-  https://e-manifest.18f.gov/api/0.1/manifest/signByTrackingNumber
+  https://e-manifest.18f.gov/api/v1/:manifest_tracking_number/sign
```
pkarman: `signature` ?

pkarman:
just some wordsmithing nits about URL syntax. 

I like splitting out `id` vs `manifest_tracking_number` as named params so that it is un-ambiguous what the request is operating on.

_static/api-examples.md
jessieay: I thought about re-naming `authenticate` to `sessions` -- but was worried that was too far from what we have already. 
What do you think of that?

The reason for plurals is that it is the Rails-y way: http://guides.rubyonrails.org/routing.html#crud-verbs-and-actions

But we can certainly go back to singular if preferred. eg: I kept `GET manifest` singular because the path is just `manifest` rather than `manifest/:id` and I didn't want people to think that was an endpoint for a collection. Not against exceptions to the Rails rule so we can stray from it wherever it makes sense.

jessieay: another idea is to have the endpoint just be `POST /sessions` and not worry about nesting at all

_static/api-examples.md
jessieay: I like that! more typing but I will jut have to reply on autocomplete :smile_cat:

jessieay: what do you think about the use of a nested route here? I like it because it is standard for Rails, but the weird thing is that we accept _either_ an id or a tracking number there...not sure what the API best practice is for something like this. As you stated earlier, explicitness of a named param is nice for knowing intent, but it's also nice to have nested routes so you know when you are acting on an object.

jessieay:
@pkarman thanks for the feedback! that is exactly the kind of input I was hoping to get on this. 

Let me know what you think about my most recent comments- I'm heading to lunch now but will be back in an hour to incorporate edits

_static/api-examples.md
pkarman: If we follow Postel's law (https://en.wikipedia.org/wiki/Robustness_principle) then I would probably make the API look like:

```
POST .../manifest/:id_or_tracking_number
```

and then have the server-side code dis-ambiguate it. That's one reason to use a UUID for the e-Manifest ID, so that it is easier to disambiguate. If we made the UUID follow a template like `em<random30charstring>` then it would be straightforward because afaik, manifest tracking numbers are always numeric.

_static/api-examples.md
pkarman: The Rails way assumes everything is CRUD and so the naming conventions map to that model. We have some CRUD but we also have some non-CRUD stuff (like signatures and authentication) so I don't feel like we need do make our API conform to Rails (rather than the other way around).

I.e. I'm trying to think about the API as a user of it, not as a implementer of a particular technology.

pkarman: Since the important thing about authenticating isn't the 'session' per se, but the actual response to be used for creating a signature, how about something like:

```
POST .../token
```

?

jessieay: great idea

_static/api-examples.md
jessieay: I like that but there's a risk in this situation -- we could have a manifest with an id of 1 and a different manifest with a tracking id of 1. What if we want the latter? How would we indicate that in the above scenario? (`POST .../manifest/:id_or_tracking_number`)

pkarman: that's why I suggested we adopt a e-Manifest id that uses UUID, so there would be little to zero risk.

jessieay: ok, let's leave this as-is for now and I will do that next!

_static/api-examples.md
jessieay: renamed at `138677410a00290bb3a7222028a0cb1135f5ef4f`

_static/api-examples.md
jessieay: renamed at c3adc27b9beecfcc70e14818cd27f65f2c423f18

jessieay:
@pkarman this should be ready for re-review. Next on my list is to add `uuid` to manifests and use that as the primary key rather than `id` or `tracking_number`

pkarman:
I would recommend using the existing integer `id` as the primary key, and add an additional unique column for `uuid` which is what gets publicly exposed. Could auto-generate that uuid value like:

```
"em#{SecureRandom.hex[0..29]}"
```

which should give us an immediately recognizable fingerprint specific to our system (in case some day manifests use another 32-character string for an identifier).

jessieay:
@pkarman separately from this PR or before merging?

why keep current id as well as having uuid? 

I was thinking we could use postgres uuid field type and use that as the primary key http://nandovieira.com/using-uuid-with-postgresql-and-activerecord

pkarman:
separate PR. That's a pretty invasive change.

I suggested 2 columns instead of one because when you use a UUID column as your primary key, then `ActiveRecord::Base.find_in_batches` doesn't work, and that's what ES uses for indexing.

jessieay:
@pkarman good to know re: elastic search!

jessieay:
maybe we should do v0 insteaed https://github.com/18F/api-standards/issues/5#issuecomment-159009677

what do you think?

pkarman:
`v0` sounds great.

jessieay:
@pkarman this is ready for re-review

_static/api-examples.md
```diff
@@ -166,7 +166,7 @@ You may also sign with the Manifest Tracking Number:
 ```bash
 curl -i -X POST -H 'Content-Type: application/json' \
   --data @sign-manifest-manifesttrackingnumber.json \
-  https://e-manifest.18f.gov/api/0.1/manifest/signByTrackingNumber
+  https://e-manifest.18f.gov/api/v0/:manifest_tracking_number/signature
```
pkarman: did we lose a `/manifests/` path part here?

jessieay: argh! looks like we did

jessieay: updated at fe3477afc5d021415db8995815f102644abac3e6

spec/requests/api/v0/post_manifest_spec.rb
```diff
@@ -1,14 +1,14 @@
 require_relative "../../../support/request_spec_helper"
 
-describe 'POST manifest/submit/:manifest_tracking_number' do
+describe 'POST manifests' do
   it 'creates a new manifest' do
     expect {
-      send_json(:post, '/api/0.1/manifest/submit/30221', { hello: 'world' })
+      send_json(:post, '/api/v1/manifests?tracking_number=30221', { hello: 'world' })
```
pkarman: still says `v1` ?

spec/requests/api/v0/post_manifest_spec.rb
```diff
@@ -1,14 +1,14 @@
...
     }.to change { Manifest.count }.by(1)
 
     manifest = Manifest.last
 
     expect(manifest.content).to eq({ 'hello' => 'world' })
-    expect(last_response.headers['Location']).to eq("/api/0.1/manifest/id/#{manifest.id}")
+    expect(last_response.headers['Location']).to eq("/api/v1/manifests/#{manifest.id}")
```
pkarman: still says `v1`

pkarman:
@jessieay the tests still refer to `v1` in lots of places.

spec/requests/api/v0/post_manifest_spec.rb
jessieay: updated at de7fa81232cdaa00608ca70b1ef2e4dc5da5cd1d

jessieay:
@pkarman argh not sure how I missed those in global find and replace. updated at de7fa81232cdaa00608ca70b1ef2e4dc5da5cd1d

PR #28, opened 2016-01-08
Sign or upload to match comps

jeremiak:
According to the [design comps](https://gsa.invisionapp.com/share/4Z4X4ALYB#/screens/115178939), the user should either sign the manifest or upload a scanned version of the manifest.

This PR makes the views capable of this sort of user flow, though currently there is no validation. That is to say that the user could just as easily not submit a scanned version with this version of the PR.

However, this sets up the UI so that the flow is consistent with the design comps.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@
 - To build and serve the app, run `rake serve`.
 - To deploy to 18F's cloud, run `rake deploy`.
 - To add dummy data for developing against, run `rake populate:manifests`.
-- To (re)build the Elasticsearch index, run `rake search:index`.
+- To (re)build the Elasticsearch index, run `rake search:index FORCE=y`.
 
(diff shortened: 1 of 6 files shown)


pkarman:
@jeremiak some merge conflicts here. My guess is whitespace in the app.js file. I tried to standardize on 2 space indent in #14 since _most_ of the code was already that way.

jeremiak:
Merge conflicts fixed (was partially whitespace) but tests that came in from `master` aren't passing for some reason now

pkarman:
@jeremiak I will investigate test failures.

pkarman:
tests are all passing for me locally. I can't figure out how to trigger travis rebuild manually, so going to make a trivial change to the branch to trigger.

pkarman:
The problem is that `rspec-mocks 3.4.1` breaks the `receive` behavior of previous versions. Working on fix now.

PR #29, opened 2016-01-08
switch to elasticsearch-rails-ha gem

pkarman:
follow on to #14


jessieay:
cool gem! looks like some pretty smart people wrote it :)

PR #30, opened 2016-01-08
Start es server before search specs only

jessieay:
- Prevents need for env var when running single specs


pkarman:
I'm a bit torn about this one because I suspect our search server will end up being used for things besides "search" but I understand the desire to speed up local tests during development.

PR #32, opened 2016-01-11
add redis config for cloud.gov

pkarman:
Story: https://trello.com/c/Xdj2N7wd

What I don't know is whether the load order of `config/redis.rb` will set the env before or after Sidekiq gets configured. This may take some try-and-see efforts at cloud.gov.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web: bundle exec rackup
-worker: bundle exec sidekiq -C config/sidekiq.yml -r ./lib/app.rb
+#worker: bundle exec sidekiq -C config/sidekiq.yml -r ./lib/app.rb
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-web: bundle exec rackup
+web: bundle exec rackup -p $PORT
 #worker: bundle exec sidekiq -C config/sidekiq.yml -r ./lib/app.rb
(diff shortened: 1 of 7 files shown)


brandocalrissian:
I don't see how we'd currently match on the ENV["RACK_ENV"] == "production" test on our cloud.gov system. The redis config seems to be set in the environment.

Do we want to set the RACK_ENV in the cloud.gov environment? I didn't see any programmatic setting for the production case either although I may have missed it.

pkarman:
We need some way to determine that the app is running at cloud.gov. All the files in `config/*rb` right now check that env var, since it is used elsewhere to determine test-vs-not-test.

So yes, I think we should set it at cloud.gov, unless there's some other way of detecting that environment that I've missed.

pkarman:
or, how is `Sinatra::Application.environment` being set?

brandocalrissian:
I don't think we set it anywhere for production; at least I can't find it.

http://www.sinatrarb.com/configuration.html

> :environment - configuration/deployment environment
> A symbol specifying the deployment environment; typically set to one of :development, :test, or :production. The :environment defaults to the value of the RACK_ENV environment variable (ENV['RACK_ENV']), or :development when no RACK_ENV environment variable is set.

I say we add it to the cloud.gov environment.

pkarman:
Cool. Done.

pkarman:
I have created a `sinatra-stable` branch with the relevant code from here. Now that we are full steam ahead on Rails, I will rework this against master.

PR #38, opened 2016-01-11
Add `manifests#new` form to Rails app

jessieay:
- Hits API endpoint

screen shot 2016-01-11 at 3 40 46 pm

response: 

screen shot 2016-01-11 at 3 40 51 pm


pkarman:
sweet!

PR #40, opened 2016-01-12
Cloud dot gov

pkarman:
I need to test this actually works as expected, once Rails app (master) is deploy-able to cloud.gov. Implements config for separate web and worker instances, for scaling purposes.


jessieay:
let us know when it's ready for review! would be good to replace deploy instructions in CONTRIBUTING w these scripts

pkarman:
feel free to review any time. I don't expect much to change, except specific config options. The concept is worth reviewing now.

script/deploy
```diff
@@ -0,0 +1,18 @@
...
+  echo "Error: appname is required.\n$USAGE" >&2
+  exit 1
+fi
+
+# http://stackoverflow.com/a/677212/358804
+if ! hash cf-blue-green 2>/dev/null; then
+  npm install -g cf-blue-green
+fi
+
+cf-blue-green $1 -f web-manifest.yml
```
jmcarp: AFAIK, `cf-blue-green` doesn't seem to support the `-f` flag. From the docs:

> cf-blue-green creates a temporary manifest from your live application, meaning that it ignores the manifest.yml in your directory, if you have one. To deploy any changes to your manifest, use cf push directly.

For that reason, the docs recommend using a different blue-green helper if possible. https://github.com/concourse/autopilot and https://github.com/bluemixgaragelondon/cf-blue-green-deploy both look decent (we're using autopilot for parts of FEC), and have the advantages of being CF CLI plugins with unit tests.

pkarman: excellent suggestion, @jmcarp, thanks. fixed in a56e880

pkarman:
If the code looks ok, this can be merged since it shouldn't interfere with anything app-specific. It's only cloud.gov deploy scripts, which we can iterate on once we can test in a dev env.

config/puma.rb
```diff
@@ -10,7 +10,8 @@
 preload_app!
 
 rackup DefaultRackup
-environment ENV.fetch("RACK_ENV", "development")
+port(ENV['PORT'] || 3000)
```
jessieay: perhaps the default should be `5000` per foreman default? or is this something totally differnet?

script/worker-start
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -e
+set -x
+
+# make sure our pids dir exists
+mkdir -p tmp/pids
```
jessieay, line 7: if we have this here should we remove from `bin/setup`?

jessieay:
two small questions, otherwise :+1: :+1:

config/puma.rb
pkarman: thanks, changed.

script/worker-start
pkarman, line 7: this is used on cloud.gov, where we do not run `bin/setup`. It's fine to have it in both places.

PR #41, opened 2016-01-12
add redis config, move elasticsearch to initializers

pkarman:


config/initializers/redis.rb
```diff
@@ -0,0 +1,16 @@
...
+# we assume defaults are fine for test/development.
+# https://github.com/mperham/sidekiq/wiki/Using-Redis
+if ENV["RACK_ENV"] == "production"
+  vcap = ENV["VCAP_SERVICES"]
+  vcap_config = JSON.parse(vcap)
+  vcap_config.keys.each do |vcap_key|
+    if vcap_key.match(/redis\d+-swarm/)
+      redis_config = vcap_config[vcap_key][0]['credentials']
+      redis_url = "redis://#{redis_config['password']}@#{redis_config['hostname']}:#{redis_config['port']}"
+      #STDERR.puts "REDIS_URL=#{redis_url}"
```
jessieay, line 11: was this left in accidentally?

Procfile
```diff
@@ -1,2 +1,2 @@
-worker: bundle exec sidekiq -C config/sidekiq.yml -r ./lib/app.rb
+worker: bundle exec sidekiq -C config/sidekiq.yml
```
jessieay, line 1: should we move the sidekiq config while we're here?

config/initializers/redis.rb
pkarman, line 11: no, but I understand why you note it. I intended to use it for testing #40 when I got to that point. I can zap it for now if you like.

Procfile
pkarman, line 1: to where? it's analogous to `config/database.yml`.

jessieay, line 1: make sense, let's keep as-is

config/initializers/redis.rb
jessieay, line 11: nah let's just keep it

jessieay:
thanks for getting back to me on the comments! merging

PR #42, opened 2016-01-12
create get-manifest.json that references post-manifest.json

pkarman:
follow-on to #26 but using `$ref` instead of duplicated schema text.

### DIFF
--- a/app/schemas/get-manifest.json
+++ b/app/schemas/get-manifest.json
@@ -6,6 +6,11 @@
   "type": [
     "object"
   ],
+  "required": [
+    "document_id",
+    "activity_id",
+    "content"
+  ],
   "definitions" : {
     "content" : {
       "$ref" : "http://e-manifest.18f.gov/schemas/post-manifest.json#"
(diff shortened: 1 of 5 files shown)


spec/support/api_schema_matcher.rb
```diff
@@ -14,4 +14,31 @@
 
     true
   end
+
+  def build_json_schema_document_store
+    # load all the schema/*json files that do *not* contain an external $ref
```
jessieay, line 19: what if we want to include definitions in a file and reference them from other schema partials? in that case, we might schemas with `ref` that we'd want to load.

app/schemas/get-manifest.json
```diff
@@ -0,0 +1,29 @@
...
+    "object"
+  ],
+  "required": [
+    "document_id",
+    "activity_id",
+    "content"
+  ],
+  "definitions" : {
+    "content" : {
+      "$ref" : "http://e-manifest.18f.gov/schemas/post-manifest.json#"
```
jessieay, line 16: what about naming the `content` file something like `manifest-content.json` and then including it in `get-manifest` and `post-manifest` ? Not sure how duplicate keys (eg: `method`) work when we ref like this.

jessieay:
going to merge this in for now, we can iterate on it going forward

PR #43, opened 2016-01-12
Show manifests at root path

jessieay:
screen shot 2016-01-12 at 1 32 30 pm


app/controllers/manifests_controller.rb
```diff
@@ -1,4 +1,9 @@
 class ManifestsController < ApplicationController
   def new
   end
+
+  def index
+    result = HTTParty.get(api_v0_manifests_search_url, headers: { "Content-Type" =>'application/json'})
```
pkarman: why not use `Manifest.authorized_search()` instead?

jessieay:
@pkarman this is ready for review now

app/controllers/manifests_controller.rb
jessieay: updated to use `Manifest.all` -- figure we will use scope once you submit your PR against this PR!

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -1,7 +1,8 @@
 class Api::V0::ManifestsController < ApplicationController
   def search
     if !params[:q] && !params[:aq]
-      render json: {}, status: 400
+      manifests = Manifest.all
```
pkarman, line 4: this does not make sense to me. An invalid search request returns all the Manifests?

app/controllers/manifests_controller.rb
```diff
@@ -1,4 +1,8 @@
 class ManifestsController < ApplicationController
   def new
   end
+
+  def index
+    @manifests = Manifest.all
```
pkarman, line 6: This is fine for now. I have a pending PR that will change it to use `authorized_search({public: true})`

app/controllers/api/v0/manifests_controller.rb
pkarman, line 4: I suspect you changed this when you were trying to use the API via external HTTP request from the manifests controller?

pkarman:
Looks fine except for a couple comments. I'm going to merge so @jeremiak can get going, and I will revert the changes to the search API that I noted as part of a pending PR.

PR #44, opened 2016-01-12
Manifest uuid

pkarman:
Trello: https://trello.com/c/dA6hnR2L

UUID is created by postgresql on insert, following v4 UUID spec: http://blog.arkency.com/2014/10/how-to-start-using-uuid-in-activerecord-with-postgresql/

### DIFF
--- a/app/controllers/api/v0/manifests_controller.rb
+++ b/app/controllers/api/v0/manifests_controller.rb
@@ -1,10 +1,9 @@
 class Api::V0::ManifestsController < ApplicationController
   def search
     if !params[:q] && !params[:aq]
-      manifests = Manifest.all
-      render json: ActiveModel::ArraySerializer.new(manifests, each_serializer: ManifestSerializer)
+      render json: {message: 'Missing q or aq param'}, status: 400
     else
-      render json: Manifest.authorized_search(params).response[:hits].to_json
+      render json: ManifestSearchSerializer.new(Manifest.authorized_search(params)).to_json(root: false)
     end
   end
 
@@ -32,7 +31,7 @@ class Api::V0::ManifestsController < ApplicationController
       return
     end
 
-    render json: manifest.to_public_json
+    render json: ManifestSerializer.new(manifest).to_json(root: false)
   end
 
   def update
@@ -46,7 +45,7 @@ class Api::V0::ManifestsController < ApplicationController
       new_json = JSON.patch(manifest_content_json, patch_json);
       manifest.update_column(:content, new_json)
 
-      render json: manifest.to_public_json
+      render json: ManifestSerializer.new(manifest).to_json(root: false)
     rescue ActiveRecord::RecordNotFound => _error
       status 404
     end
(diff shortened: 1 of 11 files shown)


app/models/manifest.rb
```diff
@@ -43,4 +43,15 @@ def self.authorized_search(params, current_user=nil)
     dsl = QueryDSL.new(params: params, current_user: current_user)
     search(dsl)
   end
+
+  def as_public_json(options={})
```
jessieay: I think that with our use of `ManifestSerializer` we shouldn't need this - we can just defined `id` on the serializers as `object.uuid`

db/migrate/20160112190459_add_uuid_to_manifests.rb
```diff
@@ -0,0 +1,18 @@
+class AddUuidToManifests < ActiveRecord::Migration
+  def up
+    enable_extension 'uuid-ossp'
+    add_column :manifests, :uuid, :uuid
+
+    Manifest.all.each do |manifest|
```
jessieay, line 9: should we throw a quick class definition of `Manifest` in here for forward compatibility? http://railsguides.net/change-data-in-migrations-like-a-boss/

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -25,26 +25,18 @@ def create
...
-        manifest = Manifest.find_by!("content -> 'generator' ->> 'manifest_tracking_number' = ?", params[:tracking_number])
-      end
+      manifest = find_manifest
     rescue ActiveRecord::RecordNotFound => _error
       render json: {}, status: 404
       return
     end
 
-    render json: manifest.to_json
+    render json: manifest.to_public_json
```
jessieay: I think we should be using `ManifestSerializer.new(manifest).to_json` here

jessieay: I had that in the PR that was closed, but the serializer did make it into the Rails PR :)

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -66,4 +58,20 @@ def manifest_params
       :manifest_tracking_number,
     )
   end
+
+  def find_manifest
+    if params[:id] || params[:uuid]
+      find_manifest_by_id
+    elsif params[:tracking_number]
+      find_manifest_by_tracking_number
```
jessieay, line 66: niiice

jessieay:
a couple of comments about how we're rendering JSON for manifest - let me know what you think!

pkarman:
This is going to fail an Elasticsearch test until we merge #48

pkarman:
I enabled ES server in 0189994 just to prove that tests will pass.

spec/serializers/manifest_serializer_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+describe ManifestSerializer do
```
jessieay, line 3: nice! I am never sure how or when to unit test serializers, but this seems like a great use case

jessieay:
looks superb -- merging!

PR #45, opened 2016-01-12
Update development docs to the current state for the rails app.

brandocalrissian:
Change the setup script to point to the existing .env.example.

Add SECRET_KEY_BASE to .env.example.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,17 +9,21 @@
 0. Start postgres if it isn't automatically running: `postgres -D /usr/local/var/postgres`
 0. Install Elasticsearch and make sure it is running.
 0. Install Redis and make sure it is running.
+0. Install the [Qt implementation of Webkit](https://github.com/thoughtbot/capybara-webkit/wiki/Installing-Qt-and-compiling-capybara-webkit) as we now depend on [capybara-webkit gem](https://github.com/thoughtbot/capybara-webkit).
 0. Run `bundle install` to grab the required gems.
 0. Use `rake db:create:all` to create the databases.
 0. User `rake db:migrate` to transform the database structure to what is
    needed by the app. For test you will need to preface with env variables 
    to indicate the environment: `RACK_ENV=test rake db:migrate`
-0. Place environment-specific variables in a `.env.`*environmentname* file. The default is `.env`.
+0. Run `./bin/setup` to initialize the rails application.
+0. Edit the environment-specific variables in a `.env.`*environmentname* file. The default created by running `./bin/setup` is `.env`.
    Test-specific variables can go in `.env.test`. If you have set a postgres username and password,
    use the `DATABASE_URL` environment variable to configure it. See `.env.example` for an example.
    Default database configuration is stored in `config/database.yml`.
-0. Run `rake serve`. This will build the Jekyll site and start the Sinatra server.
-0. Go to `localhost:9292` and enjoy!
+0. Add the CROMERR account credentials (CDX_USERNAME and CDX_PASSWORD) to `.env`.
+0. Run `rake secret` and set the output as the value for `SECRET_KEY_BASE` in `.env`.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -9,17 +9,21 @@
...
+0. Run `./bin/setup` to initialize the rails application.
+0. Edit the environment-specific variables in a `.env.`*environmentname* file. The default created by running `./bin/setup` is `.env`.
    Test-specific variables can go in `.env.test`. If you have set a postgres username and password,
    use the `DATABASE_URL` environment variable to configure it. See `.env.example` for an example.
    Default database configuration is stored in `config/database.yml`.
-0. Run `rake serve`. This will build the Jekyll site and start the Sinatra server.
-0. Go to `localhost:9292` and enjoy!
+0. Add the CROMERR account credentials (CDX_USERNAME and CDX_PASSWORD) to `.env`.
+0. Run `rake secret` and set the output as the value for `SECRET_KEY_BASE` in `.env`.
+0. Run `rails s` to start the Rails application.
```
jessieay: let's suggest `foreman start` instead so the Procfile runs

CONTRIBUTING.md
```diff
@@ -41,9 +45,9 @@
 
 ### Rake Tasks
 
-- To build the static site, run `rake build`.
-- To build and serve the app, run `rake serve`.
-- To deploy to 18F's cloud, run `rake deploy`.
+- ~~To build the static site, run `rake build`.~~
```
jessieay: should we just remove this?

CONTRIBUTING.md
```diff
@@ -9,17 +9,21 @@
...
+0. Edit the environment-specific variables in a `.env.`*environmentname* file. The default created by running `./bin/setup` is `.env`.
    Test-specific variables can go in `.env.test`. If you have set a postgres username and password,
    use the `DATABASE_URL` environment variable to configure it. See `.env.example` for an example.
    Default database configuration is stored in `config/database.yml`.
-0. Run `rake serve`. This will build the Jekyll site and start the Sinatra server.
-0. Go to `localhost:9292` and enjoy!
+0. Add the CROMERR account credentials (CDX_USERNAME and CDX_PASSWORD) to `.env`.
+0. Run `rake secret` and set the output as the value for `SECRET_KEY_BASE` in `.env`.
+0. Run `rails s` to start the Rails application.
+0. Go to `http://localhost:3000` and enjoy!
```
jessieay: foreman does not default to `localhost:3000` like the standard rails server. Perhaps we should add a foreman config for PORT so we can tell ppl where to do. Otherwise, seems like it assigns randomly (mine is at 5100) https://robots.thoughtbot.com/foreman-as-process-manager

jessieay:
these changes look great! thanks for doing this. a few small comments from me

CONTRIBUTING.md
```diff
@@ -9,17 +9,21 @@
 0. Start postgres if it isn't automatically running: `postgres -D /usr/local/var/postgres`
```
jessieay, line 8: let's remove `Copy the secret.rb.template to secret.rb. See below CROMERR Signing` while we're here

jessieay:
other things I am seeing that we can remove while we're doing this:

this section (we are removing js from the app, so no client tests):

> Client tests use karma/mocha/chai:

these are done via setup script:

> Run bundle install to grab the required gems.
> Use rake db:create:all to create the databases.
> User rake db:migrate to transform the database structure to what is needed by the app. For test you will need to preface with env variables to indicate the environment: RACK_ENV=test rake db:migrate

brandocalrissian:
Do we therefore need to run the setup script anytime we'd want to update gems or the database schema?

jessieay:
you can still run those commands in a one off basis but they aren't needed
to get set up initially, which is what I think these instructions are for

jessieay:
this looks great! merging

PR #46, opened 2016-01-12
Remove unused files from _static

jessieay:
- Were used for Jekyll app


jessieay:
Left api docs in `_static` in case those are useful when we move them over to the new Rails setup

pkarman:
A bunch of the `_includes` are used in the API examples documentation.

I'm excited to zap this too, but let's wait till all the docs are pulled over to Rails.

jessieay:
ok makes sense, will close for now

PR #48, opened 2016-01-13
new 'public' manifests search feature

pkarman:
a few things:
- reorg where search-related classes are stored
- run elasticsearch test server for any rspec with `elasticsearch: true` modifier
- add `public` search param to limit search to manifests older than 90 days.


pkarman:
the codeclimate warning is probably true, but I'm not sure, given how the Elasticsearch DSL works, how to simplify it.

pkarman:
a good night's sleep resolved the blocker.

app/models/manifest.rb
```diff
@@ -1,3 +1,5 @@
+require_relative '../search/query_dsl'
```
jessieay, line 1: I might be wrong, but I think everything in `app` is autoloaded so this shouldn't be necessary

app/models/manifest.rb
```diff
@@ -56,7 +58,7 @@ def self.rebuild_index
   end
 
   def self.authorized_search(params, current_user=nil)
```
jessieay: I might call this `user` rather than `current_user` since we're in a model and models don't have access to the `current_user` var like a controller or view might

app/controllers/manifests_controller.rb
```diff
@@ -3,6 +3,6 @@ def new
   end
 
   def index
-    @manifests = Manifest.all
+    @manifests = Manifest.authorized_search({public: true}).records.to_a
```
jessieay, line 16: should we have some type of feature spec to confirm the changed behavior of this controller?

jessieay:
@pkarman re: sleep solving problems -- love it when that happens. it reinforces everything I believe and  want to believe about working too hard being counter productive. :)

app/controllers/manifests_controller.rb
pkarman, line 16: I have added both feature and unit tests for the public filter.

app/models/manifest.rb
pkarman: I changed to `user` everywhere that is not a controller.

app/models/manifest.rb
pkarman, line 1: I thought so too, but I think the load order is what breaks things. I.e., everything in `app` might get loaded eventually, but without those explicit `require` statements I got `undefined constant` errors when running tests.

pkarman:
@jessieay thanks for the review. I think I hit all your comments.

jessieay:
awesome, merging

PR #50, opened 2016-01-13
migrate api docs to rails

pkarman:
Card: https://trello.com/c/VS8JmFLx

Some tests are marked _pending_ because there need to be specific tests written for each swagger route (which we already have in other specs). But this PR was getting big so wanted to merge as-is to get doc testing framework in place.


jessieay:
this looks cool! and like a lot less code than was needed in the previous setup --woo hoo!

one question: is there any way to ensure that these docs are updated as we update our JSON schemas? I know there is a lib you linked to at some point...would implementing that now be easier, or should we wait?

pkarman:
@jessieay by "these docs" are you referring to `swagger.json` and/or the `app/views/examples/*.json` files?

Yes, my intent is to add tests for those using our JSON schema and API so that we know when the docs are out of date. That's the Work still In Process.

pkarman:
@jessieay this is all :green_heart: now

Gemfile
```diff
@@ -48,13 +50,13 @@ group :development, :test do
   gem 'factory_girl_rails'
   gem 'pry-rails'
   gem 'rspec-rails', '~> 3.3.0'
+  gem 'apivore', git: 'https://github.com/westfieldlabs/apivore.git'
```
jessieay, line 54: wondering if we should tag a certain commit when referencing these libs via git? (helpful to know at what point in git history we were working as a way of versioning)

pkarman, line 54: `Gemfile.lock` contains the checksum used yes?

spec/spec_helper.rb
```diff
@@ -21,4 +21,4 @@
   config.order = :random
 end
 
-WebMock.disable_net_connect!(allow_localhost: true)
+WebMock.disable_net_connect!(allow_localhost: true, allow: ['json-schema.org'])
```
pkarman, line 24: all our schemas refer to the v04 schema at json-schema.org. Alternative to whitelisting it would be to download locally into `/public/schemas`, which I am not opposed to but didn't necessarily want to take on the maintenance burden.

Gemfile
jessieay, line 54: yeah it does. I guess that's good enough. thought it might be nice to know when we can move back to a version rather than replying on source code, but we can do that via gemfile lock

app/handlers/markdown_handler.rb
```diff
@@ -0,0 +1,33 @@
+class MarkdownHandler
```
jessieay, line 1: handlers! interesting dir. Seems cool. How would you define a handler ?

jessieay, line 1: as in, what types of things should we be putting in here?

config/routes.rb
```diff
@@ -12,9 +16,10 @@
       end
       resources :manifests, only: [:create]
       resources :method_codes, only: [:index]
-      patch 'manifests', to: 'manifests#update'
-      get 'manifest', to: 'manifests#show'
+      patch 'manifests/:id', to: 'manifests#update'
+      put 'manifests/:id', to: 'manifests#update'
```
jessieay: why do we need `put` and `patch` ?

config/routes.rb
```diff
@@ -12,9 +16,10 @@
       end
       resources :manifests, only: [:create]
       resources :method_codes, only: [:index]
-      patch 'manifests', to: 'manifests#update'
-      get 'manifest', to: 'manifests#show'
+      patch 'manifests/:id', to: 'manifests#update'
```
jessieay: I think that if we added `:update` to line 17 that would be the same as doing this.

config/routes.rb
```diff
@@ -12,9 +16,10 @@
...
+      put 'manifests/:id', to: 'manifests#update'
       get 'manifests/search', to: 'manifests#search'
+      get 'manifests/:id', to: 'manifests#show'
```
jessieay: I think that if we added `:show` to the array on  L 17 that would be the same as doing this

app/views/api-docs/examples/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+The production base URL is **https://e-manifest.18f.gov/api/v0**
+
+Notation note: path variables are expressed as `:variable_name`. The colon is not part of the URI.
+
+## <a name="submit-manifest"></a>Submit a manifest
+
+```bash
+curl -i -X POST -H 'Content-Type: application/json' \
+  --data @manifest.json \
+  https://e-manifest.18f.gov/api/v0/manifests?tracking_number=:manifest_tracking_number
```
jessieay: I think we changed this endpoint to not use a query param (in specs, was updated to be ` patch "/api/v0/manifests/#{manifest.uuid}",` )

app/views/api-docs/examples/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+The update request uses the HTTP `PATCH` method. See [JSON Patch](http://tools.ietf.org/html/rfc6902) and
+[JSON Pointer](http://tools.ietf.org/html/rfc6901) for specification details.
+
+Just as in the [Fetch a manifest example](#fetch-manifest), you may use either the e-Manifest ID or the
+Manifest Tracking Number.
+
+```bash
+curl -i -X PATCH -H 'Content-Type: application/json-patch+json' \
+  --data @manifest-patch.json \
+  https://e-manifest.18f.gov/api/v0/manifests?id=:manifest_id
```
jessieay: I think this endpoint was changed to not use a query param (in specs, was updated to be `patch "/api/v0/manifests/#{manifest.uuid}",` )

app/views/api-docs/examples/index.html.md
jessieay: I think this example is no longer accurate. should just be `https://e-manifest.18f.gov/api/v0/manifests` with params for the `content` values

app/views/api-docs/examples/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+the URL of the submitted manifest, which will include the unique e-Manifest identifier. A successful
+submission should return a 201 status.
+
+## <a name="fetch-manifest"></a>Fetch a manifest
+
+To retrieve a previously submitted e-Manifest, you need the e-Manifest ID from
+the [Submit a manifest example](#submit-manifest).
+
+```bash
+curl -i -X GET https://e-manifest.18f.gov/api/v0/manifest?id=:manifest_id
```
jessieay: I think this should be `https://e-manifest.18f.gov/api/v0/manifest/:uuid` bc of route `get 'manifests/:id', to: 'manifests#show'`

app/views/api-docs/examples/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+If you do not know the e-Manifest ID, but you do have the Manifest Tracking Number from Line 4 of form 8700-22,
```
jessieay, line 47: no longer true?

jessieay:
this looks awesome. left a few comments on `app/views/api-docs/examples/index.html.md` because the examples there seem a little out of sync with our API changes. 

If you'd rather we move forward and update that file next, that works for me, too.

app/handlers/markdown_handler.rb
pkarman, line 1: I was totally riffing off http://www.lugolabs.com/articles/18-render-markdown-views-with-redcarpet-and-pygment-in-rails and https://www.codefellows.org/blog/how-to-create-a-markdown-friendly-blog-in-a-rails-app

I am not at all attached to the name. Is there a better convention for where to put extensions for template handlers? I suspect the name is derivative of `ActionView::Template.register_template_handler`

config/routes.rb
pkarman: oh, we don't. I got ambitious. Since we don't really do a replace-in-place like a proper PUT, I will remove it.

config/routes.rb
pkarman: thanks, yes I tried that earlier in my dev area but because I hadn't also moved the `get /manifests/search` before it, tests were failing. Fixed now in 55a2b1d

pkarman:
@jessieay sounds like I may have been too impatient with zapping all the manifest tracking number support. I will circle back on this tomorrow re: examples.

pkarman:
@jessieay I have updated the examples doc to match the current swagger implementation.

Note that I have documented things that do not yet exist. E.g. being able to reference a manifest using either uuid or tracking number, within the URI. I did that intentionally because I wanted to make it easier for the API user. If we need to revert to explicit param naming, we can, but I think we can dis-ambiguate easily enough on the server side and it makes the API cleaner IMO.

jessieay:
@pkarman merge conflicts :(

PR #51, opened 2016-01-14
User can upload file for manifest

jessieay:
- Can submit file alone or after filling in manifest details

screen shot 2016-01-13 at 4 27 10 pm


jessieay:
@jeremiak views for uploading are in this PR. Going to work on "sign" screens next

app/controllers/manifest_uploads_controller.rb
```diff
@@ -1,5 +1,46 @@
...
+    if manifest.save
+      flash[:notice] = "Upload for manifest #{manifest.tracking_number} submitted successfully."
+      redirect_to root_path
+    end
+  end
+
+  private
+
+  def find_or_initialize_manifest
+    if params[:manifest_id]
```
pkarman, line 17: We are calling this param name `id` in the API context. Does it matter if we are inconsistent?

app/controllers/sign_or_upload_controller.rb
```diff
@@ -0,0 +1,5 @@
+class SignOrUploadController < ApplicationController
+  def new
+    @manifest = Manifest.find(params[:manifest_id])
```
pkarman, line 3: same comment about param name.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def upload_path
+    if params[:manifest_id]
```
pkarman, line 3: ditto

app/controllers/manifest_uploads_controller.rb
```diff
@@ -1,5 +1,46 @@
...
+      manifest = Manifest.find(params[:manifest_id])
+      manifest.content[:image] = image_details
```
pkarman: do we only ever support one image per manifest?

`image` is a little non-descriptive for a top-level attribute in the `content`. Could we maybe call it `uploaded_file` or something else?

app/controllers/manifest_uploads_controller.rb
jessieay, line 17: this controller is web-only. If someone were to take this action via API, it would be in the context of a POST `/manifest` or PATCH `/manifests` action (either created a new manifest with an upload or updating an existing one)

So, there isn't anything to be consistent with exactly. But, if you think something should be changed here for clarity etc, let me know!

app/controllers/manifest_uploads_controller.rb
jessieay: good idea, will change key to `uplaoded_file`

in terms of "one image" - that is my understanding of the current needs. also confirmed here that there is only one manifest per image file: https://trello.com/c/djblE1XG

app/controllers/manifest_uploads_controller.rb
pkarman, line 17: I prefer `manifest_id` so fine to leave as-is.

jessieay:
updated image key in `content` to `uploaded_file` at 146682249d1cbf9faa4d8922029ce4db09524759

jessieay:
@pkarman ready for re-review

pkarman:
:+1:

PR #54, opened 2016-01-14
Validate presence of manifest tracking number

jessieay:
- This is the only required field for a manifest


jessieay:
Im having a hard time figuring out where the CC failure is - help!

screen shot 2016-01-14 at 12 00 17 pm

brandocalrissian:
From sprint planning today: we need the manifest tracking number to be submitted with an image.

jessieay:
ok, this PR now validates presence of manifest tracking # ALWAYS!

brandocalrissian:
Can we document the HTTP 422 validation error response in our swagger docs?

No other comments-- looks good. I merged the latest master to play around with it.

PR #55, opened 2016-01-14
Use the web design gem

jeremiak:
Requires the US Web Design Standards gem and refactors the CSS accordingly.


config/routes.rb
```diff
@@ -1,6 +1,6 @@
 Rails.application.routes.draw do
   root to: 'manifests#index'
-  resources :manifests, only: [:new, :create]
+  resources :manifests, only: [:new, :show]
```
jessieay: we need to add back in `create` here so the specs don't fail :) (should be `[:new, :create, :show]`)

app/views/layouts/application.html.erb
```diff
@@ -13,19 +13,30 @@
...
+        <p>United States Environmental Protection Agency</b>
+      </div>
+      <div class="usa-width-one-half text-right">
         <%= link_to 'Home', root_path %>
-        <%= link_to 'Industry', new_submission_path %>
```
jessieay, line 22: do we still want to link to the new submission path in the header somewhere?

app/views/manifests/index.html.erb
```diff
@@ -1,7 +1,175 @@
...
+
+<div class="usa-width-one-third">
+  <form class="callout">
+    <fieldset>
+      <h2>Filter manifests</h2>
+      <label class="featured" for="generator_name">Generator name</label>
+      <input id="generator_name" ng-model="data.generator.name" name="generator_name" type="text" class="ng-pristine ng-valid ng-touched">
+
+      <label class="featured" for="tsdf_name">Disposal facility name</label>
+      <input id="tsdf_name" ng-model="data.tsdf_name" name="tsdf_name" type="text" class="ng-pristine ng-untouched ng-valid">
```
jessieay: `ng` classes left in from angular days?

app/views/manifests/index.html.erb
```diff
@@ -1,7 +1,175 @@
...
+
+      <div class="usa-grid-full">
+        <div class="usa-width-one-half">
+          <label for="origin_state">Generator state</label>
+          <select name="origin_state" id="origin_state" class="usa-input-grid-tiny tiny-input">
+            <option value="">--</option>
+            <option value="AL">AL</option>
```
jessieay: instead of listing all of these here, we can take advantage of https://github.com/18F/e-manifest/blob/master/app/helpers/states_helper.rb

``` ruby
<% us_states.each do |state| %>
 <option value=<%=state[1]%>><%=state[1]%></option>
<% end%$>
```

```
```

jessieay: or better yet, in that same helper file we can define a method and use it:

``` ruby
# app/helpers/states_helper.rb

def state_option(state)
  "<option value=#{state}>#{state}</option>"
end

```

and then in here we do:

``` ruby
<% us_states.each do |state| %>
  <%= raw state_option(state[1]) %>
<% end%$>
```

app/views/manifests/show.html.erb
```diff
@@ -0,0 +1,89 @@
...
+    <dd><%= @manifest.tracking_number %></dd>
+    <dt>Generator's Name</dt>
+    <dd><%= @manifest.generator_name %></dd>
+    <dt>Mailing Address</dt>
+    <dd><%= @manifest.tracking_number %></dd>
+    <dt>Phone</dt>
+    <dd><%= @manifest.tracking_number %></dd>
+  </dl>
+
+  <h3>Transporter</h3>
```
jessieay, line 24: there could be any number of transporters. not sure if you'd rather deal with it now or later, but we should make sure this layout deals with that nicely

PR #58, opened 2016-01-14
flag upload tests for elasticsearch server, since travis-ci fails

pkarman:
surfaced in https://travis-ci.org/18F/e-manifest/builds/102445513


jessieay:
beauty

PR #60, opened 2016-01-15
Protect from forgery with null_session for Api controllers

jessieay:
- Makes sense for regular requests, but not Api requests
- http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html
- http://stackoverflow.com/questions/16258911/rails-4-authenticity-token


pkarman:
+1

Though I will be interested to see how our discussions around authn management might affect this.

Aside: glad to see you make an `ApiController` -- was thinking this morning that that could be a decent way to standardize on error messaging. I did something similar in the past here: https://github.com/popuparchive/pop-up-archive/blob/master/app/controllers/api/base_controller.rb

PR #61, opened 2016-01-15
Create tmp/pids in the bin/setup script

brandocalrissian:
We need the tmp/pids directory to exist to run the app. Either we tell the user to create it or we take care of it here.

One. Less. Step.


jessieay:
I saw that someone else said they needed this. I myself did not need to create it. Any ideas on why?

pkarman:
you need it for sidekiq only

It's also part of #40 

I am fine to merge.

jessieay:
Clarification: I do have a `tmp/pids` dir, but I did not manually create it

PR #62, opened 2016-01-15
Remove unused `_static` dir

jessieay:
- Left over from Jekyll days


jessieay:
and just like that, we are no longer a code climate disaster

screen shot 2016-01-15 at 9 51 30 am

PR #65, opened 2016-01-15
[WIP] Manifest tracking number unique

pkarman:


app/models/manifest.rb
```diff
@@ -5,6 +5,12 @@
 class Manifest < ActiveRecord::Base
   validates :tracking_number, presence: true
 
+  before_validation :set_tracking_number, on: :create
+
+  def set_tracking_number
```
jessieay, line 10: since we already validate against tracking number in the json, what about _not_ validating the field itself so we don't have to deal with a callback?

db/migrate/20160115181255_add_manifest_tracking_number_column.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    Manifest.all.each do |manifest|
+      tracking_number = "undefined-tracking-number-#{manifest.id}"
+      if manifest.content && manifest.content['generator'] && manifest.content['generator']['manifest_tracking_number']
+        tracking_number = manifest.content['generator']['manifest_tracking_number']
+      end
+      manifest.update_column(:tracking_number, tracking_number)
+    end
+
+    change_column :manifests, :tracking_number, :string, null: false, unique: true
```
jessieay, line 16: this migration will probably fail on prod since we have manifests that were created before the manifest tracking nuber validation. how should we handle that?

app/models/manifest.rb
pkarman, line 10: where should the column number be set then? It needs to happen before the INSERT.

db/migrate/20160115181255_add_manifest_tracking_number_column.rb
pkarman, line 16: prod is currently all dummy/test data. If there is no tracking number in the `content` the column will get a `"undefined-tracking-number-#{manifest.id}"` value which is easy to search for.

app/models/manifest.rb
```diff
@@ -12,10 +18,6 @@ def content_field(json_xpath)
     end
   end
 
-  def tracking_number
```
jessieay, line 15: if we kept this method (and renamed) we could keep the json validation for tracking number and not worry about setting tracking number field in a callback

pkarman:
ok I am going to close this and go a different direction. Just convert to `content` to `jsonb` and put the unique index on there.

app/models/manifest.rb
jessieay, line 10: oh yeah, in that case we'd also not have the postgres `null: false` and could add separately

PR #66, opened 2016-01-15
Improve setup steps

jessieay:
- Populate DB and index
- Move `search` classes to `app/models/search` to fix loading problem


pkarman:
I'm mystified by the need to put `app/search` inside `app/models/search` in order for it autoload correctly. I was just reading http://urbanautomaton.com/blog/2013/08/27/rails-autoloading-hell/ and it helped enrage but not enlighten me.

Is this true for other subdirs of `app/`? Do we need to add some `config.autoload_paths << Rails.root.join('app' ...` to `config/application.rb` ?

jessieay:
@pkarman I honestly am not really sure what's going on with the search stuff. I ran `bin/rails r 'puts ActiveSupport::Dependencies.autoload_paths'` (per http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoload-paths) and `app/search` was in there...but it was still not working. :cry:

jessieay:
@pkarman think this might be what we were seeing: https://github.com/rails/rails/issues/14382#issuecomment-37702855

jessieay:
what do you think: should we keep in `app/models/search` or go down this path http://demisx.github.io/rails/2014/03/16/auto-reload-classes-custom-directories.html

pkarman:
Putting things in `app/models/search` just so they load correctly isn't going to work well long-term, IMO. That feels like a workaround. Everything isn't a model.

That said, the `require_relative` stuff is very ugly and also a workaround. So I'm not happy with either. Perfect. :(

http://demisx.github.io/rails/2014/03/16/auto-reload-classes-custom-directories.html seems like a decent solution, if it works.

jessieay:
moved classes back to `app/search` and added `app` to autload paths.

approach appears to work! 

before:

screen shot 2016-01-15 at 2 11 09 pm

after:

screen shot 2016-01-15 at 2 12 07 pm

pkarman:
lovely. so tidy.

PR #67, opened 2016-01-15
add support for user-supplied search on manifests#index

pkarman:


spec/requests/search_spec.rb
```diff
@@ -23,6 +23,15 @@
     end
   end
 
+  describe "ui" do
```
jessieay, line 26: I think it's great that we have this test. Might be easier to find if we put it in a separate file that corresponds with the path. similar to `spec/requests/api/v0/post_manifest_spec.rb` we could have the other request specs at `spec/requests/api/v1/manifests_search_spec.rb` and then have this test in a file at `spec/requests/manifest_search_spec.rb`..or something like that. 

What do you think?

jessieay:
one spec organization question, otherwise looks :+1:

pkarman:
good idea re specs.

this was mostly a demo that led to #68. I will wait till that is merged, then open a new PR with spec reorg.

PR #69, opened 2016-01-15
Add analytics code to app

jessieay:
- In Rails prod env only
- JS copied from https://github.com/18F/e-manifest/blob/57acbd249ebf44b8289d0e116da4fa189bef84f2/_static/_layouts/template/main.html#L24
- https://trello.com/c/2apE6dE7

### DIFF
--- a/app/views/application/_analytics.html.erb
+++ b/app/views/application/_analytics.html.erb
@@ -1,0 +1,13 @@
+<% if Rails.env.production? %>
+  <script type="text/javascript">
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+     -(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+     -m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+     -})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+    ga('create', 'UA-48605964-37', 'auto');
+    ga('set', 'anonymizeIp', true);
+    ga('set', 'forceSSL', true);
+    ga('send', 'pageview');
+  </script>
+<% end %>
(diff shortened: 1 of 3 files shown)


pkarman:
lovely.

any idea if we need to also add:

```
    {% if jekyll.environment == "production" %}
    <script src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js" id="_fed_an_ua_tag"></script>
    {% endif %}
```

to the footer? it's in the old `_static/_layouts/template/main.html` footer.

jeremiak:
@pkarman we absolutely do want to add DAP (in addition to the other Google Analytics) onto the site per the [18F analytics standards](https://github.com/18F/analytics-standards#digital-analytics-program-snippet)

jessieay:
good call! added at fbb35ac5589b1acf6d86e13501b3e2519acaafd3

pkarman:
@jessie: According to the DAP page:

> Note: The above _script_ tag is loaded synchronously, so you should put it at the bottom of your 
> HTML. DAP has an outstanding bug affecting asynchronous usage.

jeremiak:
@pkarman am I missing something? It looks like this analytics snippet is at the bottom of the body in the HTML document. The only time it isn't is when the `env` is `test`, in which case the analytics snippet won't be loaded anyway.

jessieay:
Yes, I believe this is at the bottom unless I too am missing something. I inserted it where it was previously (below GA code).

pkarman:
Ah then I am confused. Sorry. In the Sinatra version, the GA code was in the _head_ and the DAP code was in the footer (end of body).

jessieay:
@pkarman moved stuff around 425d7c169775d0dbc4b99dc2b9fa0cb374be1a64

jessieay:
you were right, GA recommends analytics to be before closing head tag: 

> Paste your snippet (unaltered, in its entirety) into every web page you want to track. Paste it immediately before the closing </head> tag

https://support.google.com/analytics/answer/1008080?hl=en

pkarman:
:+1:

PR #70, opened 2016-01-15
tracking number unique index

pkarman:
Card: https://trello.com/c/smpREeOM

This adds:
- converts manifests.content to `jsonb` column type
- unique index on manifest tracking number
- allows for referencing a manifest URL by either UUID or tracking number, w/o the user needing to specify which one is used.


spec/requests/api/v0/get_manifest_spec.rb
```diff
@@ -14,6 +14,17 @@
         expect(response).to match_response_schema("get-manifest")
       end
 
+      it 'finds by manifest tracking number' do
+        manifest = create(:manifest, activity_id: 1, document_id: 2)
+        manifest.reload
+
+        get "/api/v0/manifests/#{manifest.tracking_number}",
```
jessieay, line 21: do you think it makes sense to remove the ability to GET a manifest by UUID? Not sure why someone would prefer to find via UUID rather than tracking number

spec/requests/api_spec.rb
```diff
@@ -38,6 +38,42 @@
...
+            foo: ['bar', 'baz', 'quux'],
+            nested: { something: 'good' }
+          }
+        )
+
+        patch "/api/v0/manifests/#{manifest.tracking_number}",
+          patch_command.to_json,
+          set_headers
+
+        updatedManifest = Manifest.find(manifest.id)
```
jessieay: `updated_manifest` would be more rubyish

db/migrate/20160115204128_manifest_tracking_number_unique_index.rb
```diff
@@ -0,0 +1,11 @@
+class ManifestTrackingNumberUniqueIndex < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+      create unique index manifest_tracking_number_idx on manifests ((content->'generator'->>'manifest_tracking_number'));
```
jessieay, line 4: cool!

spec/requests/api_spec.rb
```diff
@@ -38,6 +38,42 @@
         expect(parsed_content).to eq(expected_content_hash)
         expect(parsed_response["id"]).to eq(manifest.uuid)
       end
+
+      it 'updates removes and adds fields to a manifest by tracking number' do
```
jessieay: put this in a request spec specific to PATCHing manifests?

jessieay:
everything related to uniqueness of manifest tracking number in here looks A++

wondering if you have thoughts about the fact that we now have two endpoints ways to hit the GET and PATCH endpoints (with UUID or manifest tracking number). I am not super opposed to both but not sure that we actually _need_ both. what do you think?

spec/requests/api/v0/get_manifest_spec.rb
pkarman, line 21: If I was using this API, I would use the UUID, especially if iterating over search results, since the `id` is a top-level attribute.

IMO good API design follows Postel's law (https://en.wikipedia.org/wiki/Robustness_principle).

spec/requests/api_spec.rb
pkarman: agreed. that code was copied from the test above it. renamed now.

spec/requests/api_spec.rb
pkarman: great. done.

pkarman:
I like having both. I don't know what the future holds for tracking number, but our UUID should be future proof.

PR #68, opened 2016-01-15
Refactor form to use search controller

jeremiak:
Modifies the form on the manifest listing to search with the values of the form.

To close https://trello.com/c/El5kKAkI/312-finish-search-filters-date-state-province-additional

PR #71, opened 2016-01-15
re-org tests to match URLs

pkarman:
follow on to #68 to add tests for `/manifests` search


jessieay:
specs aren't liking `require_relative`

db/chores/populator.rb
```diff
@@ -28,147 +28,7 @@ def make_manifest_record(iteration)
...
-  "waste_non_hazardous_material_present": "yes",
-  "waste_handling_instructions": "Do not feed after midnight.",
-  "report_management_method_codes": [
-    "Method1",
-    "Method2",
-    "Method3"
-  ]
-}
-    JSON
+    File.read(Rails.root.join('app', 'views', 'examples', '_manifest.json'))
```
jessieay, line 31: oh wow, didn't know we had this file. What about putting it in `spec/support/fixtures`?  or maybe just `db/chores/data` ? (new dir)

spec/requests/api/v0/manifests_search_spec.rb
```diff
@@ -1,35 +1,31 @@
 require 'rails_helper'
-require_relative '../../db/chores/populator'
 require "queryparams"
 
 describe "search", elasticsearch: true do
+  before(:each) do
+    Manifest.rebuild_index
```
jessieay: I'd probably prefer this to be inlined. If it essential to a test passing correctly, having it inline tells me what the dependencies of the test are.

spec/requests/api/v0/manifests_search_spec.rb
```diff
@@ -1,35 +1,31 @@
...
+  end
+
   describe "/search" do
     it "searches all fields by default" do
-      populate_manifests(2)
-      search_for "around+the+corner"
+      2.times.map { create(:manifest, :indexed) }
```
jessieay, line 7: nice! another shorthand for this (from FactoryGirl) is `create_list(:manifest, 2, :indexed)` (not really that much shorter, though. I am cool with either)

spec/requests/manifests_search_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+require "queryparams"
+
+describe 'manifests search', elasticsearch: true do
+  before(:each) do
+    Manifest.rebuild_index
```
jessieay: inline?

jessieay:
few small Qs, otherwise looks ready to go!

db/chores/populator.rb
pkarman, line 31: the examples are used in the documentation, so they need to sit in `app/views` for partials to work.

spec/requests/api/v0/manifests_search_spec.rb
pkarman: it's a bit like database cleaner: resetting the storage to a known good state. Would you like to see that inlined as well?

jessieay: I agree that this is similar to database cleaner, but it is not exactly the same. If I write a new test, I know that I will always want to a assume an empty DB. I am less sure that the state of the manifest index is relevant (but it might be!)

Will `Manifest.rebuild_index` be required for all `elasticsearch: true` specs? And if so, what about including it in that code somehow?

pkarman:
@jessieay good call about rebuild_index being part of the helper setup. I have moved it there in 2d2c8c8

jessieay:
changes are :+1: -- mergin'

PR #75, opened 2016-01-19
add validation endpoint

pkarman:
Cards:
- https://trello.com/c/YtLBPfLy
- https://trello.com/c/O2Qu6Q2J

This PR also includes fixes to the post-manifest schema and example docs.


config/routes.rb
```diff
@@ -18,6 +18,7 @@
     namespace :v0 do
       resources :tokens, only: [:create]
       get 'manifests/search', to: 'manifests#search'
+      post 'manifests/validate', to: 'manifests#validate'
```
jessieay, line 21: what do you think about an endpoint at `manifests/validations#create` instead? that way we are working within the standard rails endpoints

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -9,7 +9,34 @@ def search
...
+  end
+
   def create
+    validator = ManifestValidator.new(manifest_params)
+    unless validator.run
+      render json: {
+        message: "Validation failed",
+        errors: validator.error_messages
+      }.to_json, status: 422
+      return
```
jessieay: do we need this `return` ?

app/services/manifest_validator.rb
```diff
@@ -0,0 +1,28 @@
+class ManifestValidator
+  include JsonSchemaHelper
+
+  attr_reader :content, :errors
```
jessieay: do we need this `attr_reader` to be public? looks like we are only using these ivars internally

app/views/api-docs/examples/index.html.md
```diff
@@ -35,6 +35,21 @@ Status: 201
 Location: https://e-manifest.18f.gov/api/v0/manifests/de305d54-75b4-431b-adb2-eb6b9e546014
 ```
 
+## <a name="validate-manifest"></a>Valudate a manifest
+
+Before you submit a manifest, you may validate its content and structure.
```
jessieay, line 40: this is so cool!

app/controllers/api/v0/manifests_controller.rb
jessieay: ahh just viewed whole file and now I see why we have this.

what about doing an `if` and `else` in this `create` method rather than an `unless` with a return. I think that might make the logic easier to understand

spec/services/manifest_validator_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+describe ManifestValidator do
+  it "handles either JSON string or object" do
+    json = { foo: 'bar' }
+    validator = ManifestValidator.new(json)
+    expect(validator.content).to eq( json )
+    validator = ManifestValidator.new(json.to_json)
+    expect(validator.content).to eq( { 'foo' => 'bar' } )
+  end
+
+  it "returns false on invalid JSON" do
```
jessieay, line 12: should we have a test for valid json, too?

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -9,7 +9,34 @@ def search
     end
   end
 
+  def validate
+    begin
+      manifest_content = JSON.parse(request.body.read)
+    rescue JSON::ParserError => err
```
jessieay: what about `error` instead of `err`?

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -9,7 +9,34 @@ def search
...
+      render json: {message: "Invalid JSON in request: #{err}"}, status: 400
```
jessieay: perhaps the validator service object should take care of this logic?

app/services/manifest_validator.rb
```diff
@@ -0,0 +1,28 @@
...
+    else
+      @content = content
+    end
+    register_schemas_by_uri
+  end
+
+  def run
+    schema_file_path = schema_file('post-manifest')
+    @errors = JSON::Validator.fully_validate(schema_file_path, content, errors_as_objects: true)
+    !@errors.any?
```
jessieay, line 18: what if run returned a hash:

``` ruby
{
  errors: true,
  messages: "blah blah"
}
```

then, when using this service class, we'd access both the error and potential messages at the same time. Not sure this implementation is better than what we have here, but something to consider

public/api-documentation/swagger.json
```diff
@@ -151,6 +151,40 @@
...
+                    }
+                ],
+                "tags": [
+                    "manifests"
+                ],
+                "responses": {
+                    "201": {
+                        "description": "Successful validation of e-Manifest."
+                    },
+                    "422": {
```
jessieay, line 176: do we include what the response body would be for this response somewhere in the docs? That would be very helpful

jessieay:
did a round of review - back to you, pkarman!

config/routes.rb
pkarman, line 21: there is no such object as a `validation` to create. This service is analogous to `search`: it's something you do, not something you act upon. I.e. it's not REST-style, it's RPC-style.

app/services/manifest_validator.rb
pkarman: they are exercised in the unit tests, which validate that the internals work as expected. I.e. that you can pass a String or Hash into the constructor, and get the full error stack back out after `run`.

spec/services/manifest_validator_spec.rb
pkarman, line 12: added

app/controllers/api/v0/manifests_controller.rb
pkarman: renamed

app/controllers/api/v0/manifests_controller.rb
pkarman: the controller/view logic should be handled by the validator?

app/services/manifest_validator.rb
jessieay: I'd prefer to test public facing methods only. the smaller the API of our objects, the better! if we think there is logic happening in private methods that is complicated enough to warrant its own test, perhaps it's time to extract that into another object

app/services/manifest_validator.rb
pkarman, line 18: I toyed with these 2 interfaces:

``` ruby
# exception-based
begin
  validator.run
rescue ManifestValidationError => error
  puts "You had problems: #{error.messages("\n")}"
end
# or return code-based
unless validator.run
  puts "You had problems: #{validator.error_messages.join("\n")}"
end
```

Both of those seem consistent with how Ruby/Rails handles errors. I kind of prefer the former, since catastrophic errors are your friend, but it seemed less verbose to have `run` return a simple boolean.

pkarman, line 18: All of that is to say, returning a special hash means your code would have to look like:

``` ruby
response = validator.run
if response[:errors] == true
  puts "You had problems: #{response.messages.join("\n")}"
end
```

which seems like it is both verbose and prone to error (what if you typo the hash key and check for `:errorz` instead? you might never know you had any problems).

public/api-documentation/swagger.json
pkarman, line 176: I have a TODO for that based on https://trello.com/c/qULisG3d

config/routes.rb
jessieay, line 21: it's not an object in our db, but it is a resource. 

similar to a `token` for which we to `tokens#create` (there are no tokens in the db)

You could argue that `search` should live in its own `SearchesController` as well so it would fall into a standard routing structure.

pkarman, line 21: What does a Validation look like? If we are going to make Validations (rather than just Errors), then I would be happy to promote this endpoint to a first-class controller. What would you do with a Validation, besides report it back to the user?

I would rather see `GET /manifests` route to `manifests#search` rather than have a whole controller.

By "standard" routing structure do you mean CRUD-like Rails routing? I consider that one standard among others.

jessieay, line 21: Yes sorry for lack of clarity: by standard, I mean Rails standard. Since we are using Rails, I think it makes sense to stick with Rails standards as long as they make sense. We seem to be using Rails standard routing for this app, so I am trying to figure out the best way to stick with that standard throughout _all_ of our routes.

The way I think about it in this case is:
- when we are validating a manifest, we are POSTing some data that does not update an existing object, so I think `create` is the most accurate description of this action in the context of the 7 rails default actions (GET collection/new/single object/edit, POST create, PATCH update, DELETE destroy). 
- if we agree on above, we should agree on a name of _what_ is being created. It is not a manifest, certainly. It is a validation. It is a validation of a manifest, though, so I like the idea of nesting the route to make that clearer
- based on 2 bullets above, I think a logical place to put this would be `manifests/validations#create` (`create` action in a nested controller named `ValidationsController`)

What we have no works just as well, but as a precedent I think it makes sense to try and include new controller actions into one of the 7 rails defaults. There will be exceptions to this, no doubt, but I don't think this case is unusual enough to require an exception.

public/api-documentation/swagger.json
jessieay, line 176: :+1:

config/routes.rb
pkarman, line 21: When you create a validation, what would you expect to be returned? What is this Validation thing that you, as an API user, are creating? Where does this Validation, once created, exist?

I.e., is there a Validation as a Noun, not a Verb?

Rails routing is biased toward Resourceful CRUD, but that is not all Rails supports (http://guides.rubyonrails.org/routing.html#non-resourceful-routes). As such, this doesn't feel like an exception. It just isn't a Resource CRUD action, imo.

It seems to me you are shoehorning an action (validating) into a Resourceful shoe, and it just doesn't fit.

pkarman, line 21: how about instead:

```
POST /manifests?validate=1
```

app/services/manifest_validator.rb
pkarman: I have removed the `content` accessor in 34ff86f. The `errors` accessor is still valid, since it returns the entire stack of error objects, not just the message strings.

config/routes.rb
jessieay, line 21: I see your point about validation being a verb not a noun. I think it can go either way, but fine to keep as-is. 

What does the 1 represent at `POST /manifests?validate=1` ?

pkarman, line 21: oh, the `1` is just a true value.

If we could have arbitrary HTTP verbs, this would be ideal:

`VALIDATE /manifests  < manifest.json`

but alas, we are constrained.

PR #77, opened 2016-01-20
Add login form for Crommerr signing

jessieay:
- Not tested at feature level because stubbing SOAP API proved difficult

screen shot 2016-01-19 at 4 35 02 pm

https://trello.com/c/HmXIrs7B

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -136,8 +136,6 @@ INSERT INTO schema_migrations (version) VALUES ('20150924181500');
 
 INSERT INTO schema_migrations (version) VALUES ('20160112190459');
 
-INSERT INTO schema_migrations (version) VALUES ('20160115181255');
-
 INSERT INTO schema_migrations (version) VALUES ('20160115203719');
 
 INSERT INTO schema_migrations (version) VALUES ('20160115204128');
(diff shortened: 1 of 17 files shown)


db/structure.sql
```diff
@@ -136,8 +136,6 @@ INSERT INTO schema_migrations (version) VALUES ('20150924181500');
 
 INSERT INTO schema_migrations (version) VALUES ('20160112190459');
 
-INSERT INTO schema_migrations (version) VALUES ('20160115181255');
```
jessieay, line 139: hmmm not sure why this was removed. will git checkout

spec/support/es_spec_helper.rb
```diff
@@ -56,6 +56,8 @@ def create_es_index(klass)
     create_es_index(Manifest)
   end
   config.after :suite do
-    stop_es_server
+    if es_server_running?
```
pkarman: the check for `es_server_running?` already happens inside `stop_es_server` -- any reason you needed to check 2x?

pkarman:
@jessieay just a minor comment. Otherwise looks great. Would be great if there was some way to reliably test/mock the SOAP stuff. Alas.

spec/support/es_spec_helper.rb
jessieay: ah yes - this as an accidental re-add. removed webmock to try and test the actual CDX API responses, which was causing es problems. removed at c2a6d24

jessieay:
@pkarman this is ready for re-review

PR #79, opened 2016-01-20
Refactor CDX classes

jessieay:
- Consistently use new hash syntax and symbols for hash keys
- Do not use aliased methods (hard to grep for)

### DIFF
--- a/app/models/cdx/activity.rb
+++ b/app/models/cdx/activity.rb
@@ -1,24 +1,35 @@
 class CDX::Activity < CDX::LoggedRequest
-  alias :create :perform
-
   private
+
+  def request
+    client.call(
+      :create_activity_with_properties,
+      {
+        message: {
+          securityToken: opts[:token],
+          signatureUser: opts[:signature_user],
+          dataflowName: opts[:dataflow_name],
+          properties: request_properties
+        }
+      }
+    )
+  end
 
   def request_properties
     [
-      {:Property => {:Key => "activityDescription", :Value => opts[:activity_description]}},
-      {:Property => {:Key => "roleCode", :Value => opts[:role_code]}}
-    ]
-  end
-
-  def request
-    client.call(:create_activity_with_properties, {
-      message: {
-        :securityToken => opts[:token],
-        :signatureUser => opts[:signature_user],
-        :dataflowName => opts[:dataflow_name],
-        :properties => request_properties
+      {
+        Property: {
+          Key: 'activityDescription',
+          Value: opts[:activity_description]
+        }
+      },
+      {
+        Property: {
+          Key: 'roleCode',
+          Value: opts[:role_code]
+        }
       }
-    })
+    ]
   end
 
   def repackage_response
--- a/app/models/cdx/client.rb
+++ b/app/models/cdx/client.rb
@@ -1,13 +1,15 @@
 class CDX::Client
   attr_reader :savon
 
+  CDX_BASE_URL = ENV['CDX_BASE_URL'] || "https://devngn.epacdxnode.net"
+
   def initialize(opts={})
     @savon = Savon.client(default_opts.merge(opts))
   end
 
   def default_opts
     {
-      wsdl: "https://devngn.epacdxnode.net/cdx-register/services/RegisterSignService?wsdl",
+      wsdl: "#{CDX_BASE_URL}/cdx-register/services/RegisterSignService?wsdl",
       pretty_print_xml: true,
       log: true,
       soap_version: 2,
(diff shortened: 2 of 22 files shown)


jessieay:
@brandocalrissian would love your feedback on this, since you're most familiar with CDX. I was also interested in converting our hash keys to snake_case when possible but it seems like in some cases we are sending camelcase keys because that is what the CDX API wants.

Suggestions on how I can figure out where the hash keys we're sending are required by CDX and where they are only used internally? Test coverage seems good but there is a lot of stubbing going on so hard to tell where a change in implementation would break a non-stubbed environment.

pkarman:
@jessieay I'm +1 on the syntax changes but I will be no help on the functional review, since I don't know anything about CDX and there do not seem to be any functional tests.

Maybe we should re-enable the functional tests in spec to bypass webmock if a particular env var is set?

jessieay:
@pkarman just opened a PR using VCR to record the actual API responses from CDX.  https://github.com/18F/e-manifest/pull/82

Perhaps I can expand on that in a separate PR. Then we can remove all of the mocking and stubbing of our internal CDX classes in request specs and use the recorded API responses instead. So similar to the idea of bypassing webmock except we will record the responses and use those instead of hitting the CDX API repeatedly (which I did a few times once and got us locked out of our account..oops).

What do you think?

pkarman:
+1 to VCR. I like having actual lifelike responses to test against. Like madame tussauds but for code. ;)

pkarman:
@jessieay in conflict now. Maybe can now use VCR stuff since it is in master?

jessieay:
@pkarman this updates about half of the CDX classes to use VCR instead of mocking / stubbing internal objects. Figure it will be easier to review this in two parts since it's a bit large.

app/models/cdx/activity.rb
```diff
@@ -1,24 +1,35 @@
 class CDX::Activity < CDX::LoggedRequest
-  alias :create :perform
```
pkarman, line 2: I don't see this alias re-implemented in some other way. Does that matter?

pkarman, line 2: Ah I see. You are calling `perform` explicitly elsewhere. Got it.

app/models/cdx/client.rb
```diff
@@ -7,25 +7,27 @@ def initialize(opts={})
 
   def default_opts
     {
-      :wsdl => "https://devngn.epacdxnode.net/cdx-register/services/RegisterSignService?wsdl",
-      :pretty_print_xml => true,
-      :log => true,
-      :soap_version => 2,
-      :convert_request_keys_to => :none,
+      wsdl: "https://devngn.epacdxnode.net/cdx-register/services/RegisterSignService?wsdl",
```
pkarman: Should this URL maybe be an env var instead? Or some config option. I am imagining it would change per-environment.

app/models/cdx/client.rb
```diff
@@ -7,25 +7,27 @@ def initialize(opts={})
...
+    )
   end
 
   def self.auth
     new({
-      :filters => [:password],
-      :wsdl => "https://devngn.epacdxnode.net/cdx-register/services/RegisterAuthService?wsdl"
+      filters: [:password],
+      wsdl: "https://devngn.epacdxnode.net/cdx-register/services/RegisterAuthService?wsdl"
     })
```
pkarman, line 33: Same thing with this URL.

app/models/cdx/question.rb
```diff
@@ -1,16 +1,17 @@
...
-        :securityToken => opts[:token],
-        :activityId =>    opts[:activity_id],
-        :userId =>        opts[:user][:UserId]
+    client.call(
+      :get_question,
+      {
+        message: {
+          securityToken: opts[:token],
+          activityId: opts[:activity_id],
+          userId: opts[:user][:UserId]
```
pkarman, line 11: should this be `:user_id` instead?

pkarman:
@jessieay looks good. Just a few clarifying questions.

app/models/cdx/activity.rb
jessieay, line 2: yes I generally prefer to avoid aliased methods because they can be hard to grep for (would tend to grep for `def create` to find this method, which would not work)

app/models/cdx/client.rb
jessieay: added URL base to env var (with default value) here: eec1fa4

What do you think of that? I assume the rest of the URL will remain the same (but that is an assumption)

app/models/cdx/question.rb
jessieay, line 11: `get_question` is a method we are passing to the SOAP client, which hits CDX. I believe CDX requires camelcase args since the XML keys are camelcase.

jessieay:
@pkarman responded to your feedback, thanks for the review!!

pkarman:
lovely.

Issue #74, opened 2016-01-19
New requirements for database migration


bbingCGI:
When updating to the latest, I ran into two problems.
1) Postgres 9.4 or higher is required for jsonb data type.  (Ubuntu 14.04 comes with 9.3)
2) rake db:migrate requires a user with permission to create extensions  (such as postgres user)

Both were easy to overcome.  But thought it was important to note.

brandocalrissian:
We need to document the postgres requirement and create the role in our setup script.

brandocalrissian:
Something like `createuser -s -r postgres`

pkarman:
just discovered cloud.gov uses 9.3 also, so the `jsonb` feature is going to need to be re-thought in any case.

pkarman:
going to create a PR to rollback that column change. The uuid extension is still required however.

Should not need to create a new user to install the extension, but should document sudo of some kind might be necessary.

jessieay:
PR #81 was merged so this issue has been addressed. Closing.

PR #81, opened 2016-01-20
revert manifests.content back to json column type

pkarman:
see discussion on #74


jessieay:
cool - let's update our travis settings to use 9.3 then, too. (currently uses 9.4)

pkarman:
@jessieay will do, one sec.

jessieay:
for those of us who have already run this migration in its old form, what will be the best way to update? `rake db:migrate` won't be problematic?

pkarman:
@jessieay I don't think it will be. If you have already run it, this will migrate it back. If you haven't run it, it will basically do nothing (just tested that at cloud.gov).

PR #82, opened 2016-01-20
Do not return token on failed auth with CDX

jessieay:
- Response status 401 instead of 200
- Use VCR to record API responses
- https://trello.com/c/yoUBP2u9


jessieay:
getting some weird failures on this PR. halp!

jessieay:
ahhh needed to ignore locahost in VCR config. should be fixed now

pkarman:
lovely. assuming codeclimate is hung indefinitely.

PR #84, opened 2016-01-20
use manifest.uuid (not .id) in links to individual manifests

pkarman:


jessieay:
this is failing CI. looks good otherwise

PR #85, opened 2016-01-20
cloud.gov fixes

pkarman:
fixes redis url and increases our memory allocation.


jessieay:
code climate hanging - argh! merging

PR #86, opened 2016-01-20
Wrap views in usa grid classes

jeremiak:
To prevent a small gap of blue between the main content and the nav content we need to wrap the views in the `usa-width-*` classes. We can’t do this at the `application.html.erb` level because we need the views to be able to configure themselves as two-thirds/one-third or full width. For example, the index search view has a right hand section for the search query form.

Blue gap example:
Example of blue gap between nav and content

PR changes:
Changes to UI with PR, no blue gap

### DIFF
--- a/app/views/manifest_uploads/new.html.erb
+++ b/app/views/manifest_uploads/new.html.erb
@@ -1,9 +1,13 @@
-<h3>Attach a scan of the manifest</h3>
+<div class="usa-width-one-whole">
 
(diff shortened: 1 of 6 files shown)


jessieay:
This is somewhat unrelated but might make sense to include here:

`app/views/tokens/new.html.erb` does not have any text to give context on what the user is signing into. Perhaps we should add some text above that says something like "Log in with your CDX credentials" ?

pkarman:
Let's change the UX text for the sign in page in a different PR. Might take some wordsmithing.

PR #88, opened 2016-01-20
use favicon from epa.gov

pkarman:


jeremiak:
Looks simple enough to me! I'll merge when green

PR #89, opened 2016-01-20
Sidekiq at cloud dot gov

pkarman:
Defer to the existing Elasticsearch client config rather than re-implementing it.


jessieay:
interesting! after all of the work with es here and on C2, I think you're around ready to write a blog post about es + Cloud.gov! :)

pkarman:
@jessieay heh, yeah no doubt. I think @melodykramer and I are talking about that soonish.

PR #90, opened 2016-01-20
Swagger tests

pkarman:
Trello: https://trello.com/c/VS8JmFLx

There are several changes to the `post-manifest.json` schema here that also touch on items in https://trello.com/c/4zYZoWWl


spec/docs/swagger_spec.rb
```diff
@@ -10,10 +10,57 @@
...
   context 'has valid paths' do
     it do
+      expect(subject).to validate(:get, '/manifests/{id}', 200, params)
+    end
+    it do
+      expect(subject).to validate(:get, '/manifests/{id}', 404, unknown_manifest_params)
+    end
+    it do
+      expect(subject).to validate(
+        # TODO must we require 'manifest' key? that is inconsistent with PATCH and /validate
```
pkarman, line 20: The `update` (PATCH) and `validate` methods do:

```
manifest_content = JSON.parse(request.body.read)
```

while the `create` (POST) method reads from params, depending on Rails to decode the JSON. This means a top-level param called `manifest`.

What that means is that the examples documentation for `create` are wrong, and inconsistent with the `validate` method (which IMO should have identical APIs except for the endpoint).

@jessieay thoughts on dropping the requirement for a top-level `manifest` param?

pkarman:
This is getting big and the swagger tests are all passing, if incomplete (TODO signature and tokens tests).

Let's review and discuss lest this PR get any bigger.

spec/docs/swagger_spec.rb
jessieay, line 20: The challenge with dropping the top level param is that we're gonna have to fight Rails if we do. 

With strong params, we need to explicitly state which param values that we allow. If we allow top-level params, Rails forms will send additional params and we will end up with this error msg:

eg: here is us allowing top level params

```
params.permit(:manifest_attr, :manifest_atr_2)
```

here is the error msg we will get:

```
ActionController::UnpermittedParameters:
       found unpermitted parameters: utf8, token, commit
```

whereas 

```
params.require(:manifest).permit(:manifest_attr_1, :manifest_attr2)
```

makes sure we only look within `manifest` so those other top-level params (utf8, token, commit) are ignored.

pkarman, line 20: why use `params` at all? Why not use `manifest_content = JSON.parse(request.body.read)` like `update` and `validate` do?

I.e. the API needs the JSON. Any other controller might use `params` (as from a form).

jessieay, line 20: We could do that. 

I've been moving our API over to use params because then the API controllers and web controllers are working more similarly. I like the pattern of having shared controller concerns for the API and Web params.

Do you think the benefit of having the API controllers work differently is worth the risk of having to update two controllers in the case that, say, we add a field to the JSON schema?

jessieay:
will merge this now, we can discuss API changes later!

PR #92, opened 2016-01-21
add emanifest_env helper method

pkarman:
Trello: https://trello.com/c/2apE6dE7

This allows us to run multiple instances on cloud.gov and designate only one as "production".


jessieay:
V smart!

PR #94, opened 2016-01-21
Updating manifest examples and schema

scottdchristian:
Pull request to update manifest schema and the examples


scottdchristian:
Pulled the wrong thing down.  Doing a pull of 90 rather than master

PR #95, opened 2016-01-21
do not store this in git

pkarman:
Here's a script to use instead:

https://github.com/pkarman/my-toolbox/blob/master/cssh


jessieay:
what is the reason for not storing the linked script in this repo?

jessieay:
(I am sure there is a good one, just wondering)

pkarman:
I have a single copy of `cssh` I store in `$HOME/bin` so I can use it across all projects/repos. I don't care if we put a copy here too. It's more of a developer tool than specific to this project though. (I know, we copied it locally for C2 -- just didn't want to get into the maintenance headache of having same code in multiple repos).

jessieay:
@pkarman makes sense!

PR #97, opened 2016-01-21
Embed swagger docs in application layout

jeremiak:
To keep the UI of the site consistent, the swagger docs are now using the application layout and are thus displayed in the same interface as the rest of the web application.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,11 +1,14 @@
 @charset "utf-8";
 
 /*
+
  * = require us_web_design_standards
  * = require us_web_design_standards_fonts
+ * = require swagger-ui/screen
  */
 
  @import "variables";
+ @import "reset";
 
  @import "callout";
  @import "disclaimer";
@@ -25,3 +28,8 @@
  main {
    background-color: $base-background-color;
  }
+
+.swagger-section ul,
+.swagger-section ol {
+  list-style-type: none;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -8,7 +8,6 @@
  */
 
  @import "variables";
- @import "reset";
 
  @import "callout";
  @import "disclaimer";
@@ -16,6 +15,7 @@
  @import "headings";
  @import "lists";
  @import "nav";
+ @import "swagger";
  @import "tables";
  @import "text";
 
@@ -28,8 +28,3 @@
  main {
    background-color: $base-background-color;
  }
-
-.swagger-section ul,
-.swagger-section ol {
-  list-style-type: none;
-}
--- a/app/controllers/api_documentation_controller.rb
+++ b/app/controllers/api_documentation_controller.rb
@@ -4,6 +4,6 @@ class ApiDocumentationController < ApplicationController
   end
 
   def swagger
-    render 'api-docs/swagger/index', layout: false
+    render 'api-docs/swagger/index'
   end
 end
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -1,11 +1,14 @@
...
 
 /*
+
  * = require us_web_design_standards
  * = require us_web_design_standards_fonts
+ * = require swagger-ui/screen
  */
 
  @import "variables";
+ @import "reset";
```
jessieay: it seems like this css file is specific to swagger styles -- maybe name it more specifically for clarity?

app/assets/stylesheets/application.scss
```diff
@@ -25,3 +28,8 @@
  main {
    background-color: $base-background-color;
  }
+
+.swagger-section ul,
```
jessieay: should these be in the other file?

app/controllers/api_documentation_controller.rb
```diff
@@ -4,6 +4,6 @@ def examples
   end
 
   def swagger
-    render 'api-docs/swagger/index', layout: false
+    render 'api-docs/swagger/index'
```
jessieay, line 7: yay!

app/views/api-docs/swagger/index.html.erb
```diff
@@ -1,24 +1,5 @@
...
-      </div>
-    </div>
-  </header>
-<%= render "javascript" %>
-<%= render 'swagger_ui/swagger_ui', discovery_url: 'swagger.json' %>
-</body>
-</html>
+<%= stylesheet_link_tag 'swagger-ui/screen' %>
+<div class="swagger-section usa-width-one-whole">
+  <%= render "javascript" %>
```
jessieay: I believe the application layout already renders javascript so this might be a dupe

pkarman: The reason that the original implementation did not use the application layout is that the `javascript` invocation in the main app layout happens at the end of the _body_ tag, and the swagger_ui partial needs jquery to already be loaded. Basically loading the javascript before the body rather than after it.

app/views/api-docs/swagger/index.html.erb
```diff
@@ -1,24 +1,5 @@
...
-        <%= link_to 'Home', root_path %>
-        <%= link_to 'Industry', new_submission_path %>
-      </div>
-    </div>
-  </header>
-<%= render "javascript" %>
-<%= render 'swagger_ui/swagger_ui', discovery_url: 'swagger.json' %>
-</body>
-</html>
+<%= stylesheet_link_tag 'swagger-ui/screen' %>
```
pkarman: I believe stylesheet tags belong in the _head_ section, not the _body_.

app/views/api-docs/swagger/index.html.erb
jessieay: ah ok - maybe we can `<%= render "swagger_javascript" %>` here instead and load jquery in that partial? Seems like we shouldn't have to re-load _all_ js twice

pkarman: it wasn't loaded twice in the original.

can we just change application layout to load javascript before the body rather than after? I have hit this problem in many apps. The desire to load js at the end of the page makes it difficult to do any inline js in the body.

app/views/api-docs/swagger/index.html.erb
jeremiak: Handled in Sass compilation, so removed from the view

pkarman:
looks good to me. @jessieay you ok with this change?

jessieay:
looks good!

PR #101, opened 2016-01-22
load test

pkarman:
Trello: https://trello.com/c/qFQhX4Fh

In addition to load testing, this PR also contains some enhancements to the CDX model logging and some fixes to signature creation.


pkarman:
~~I am inclined to ignore these codeclimate duplicate code errors, since this code exists outside the Rails app (really could be its own repo but I would rather keep it together for ease-of-use).~~

pkarman:
This PR is now branched from #102 so please merge that one first.

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -1,4 +1,8 @@
 class Api::V0::ManifestsController < ApiController
+  include Rails.application.routes.url_helpers
```
jessieay: I think these are included in rails controllers by default

app/controllers/api/v0/signatures_controller.rb
```diff
@@ -1,21 +1,38 @@
...
   def create
     manifest = find_manifest(params[:manifest_id])
-    sign_request = read_body_as_json
+    signature_request = prep_signature_request(manifest)
 
-    manifest_content = manifest[:content].to_json
-    sign_request[:manifest_content] = manifest_content
-    system_session_token = session[:system_session_token]
-    sign_request["token"] = system_session_token
+    cdx_start = Time.now
```
jessieay: `Time.current` so we can ensure UTC time? http://winstonyw.com/2014/03/03/time_now_vs_time_zone_now/

app/controllers/api/v0/signatures_controller.rb
```diff
@@ -1,21 +1,38 @@
...
 
-    render json: response.to_json unless performed?
+  def update_manifest(cdx_response, signature_request, manifest)
+    if cdx_response.key?(:document_id)
+      manifest.document_id = cdx_response[:document_id]
+      manifest.activity_id = signature_request[:activity_id]
+      manifest.save!
+      status = 200
+    else
+      status = 422
```
jessieay: I think we could return `422` here and `200` above and not worry about returning `status` at the end of this method

app/controllers/api/v0/tokens_controller.rb
```diff
@@ -2,11 +2,13 @@ class Api::V0::TokensController < ApiController
   include AuthParams
 
   def create
+    cdx_start = Time.now
```
jessieay: `Time.current`

app/controllers/api/v0/signatures_controller.rb
```diff
@@ -1,21 +1,38 @@
...
+    Rails.logger.debug(ANSI.blue{ "  CDX signature time: #{sprintf('%#g', (cdx_stop - cdx_start))} seconds" })
 
-    response = CDX::Manifest.new(sign_request).sign
+    status = update_manifest(cdx_response, signature_request, manifest)
 
-    if (response.key?(:document_id))
-      manifest[:document_id] = response[:document_id]
-      manifest[:activity_id] = sign_request["activity_id"]
-      manifest.save!
+    render(json: cdx_response.to_json, status: status) unless performed?
```
jessieay: I usually prefer `if` / `unless` statements to be indented rather than at the end of a line like this. more code, but easier to read imo.

app/controllers/api/v0/manifests_controller.rb
pkarman: good call!

app/controllers/api/v0/signatures_controller.rb
pkarman: do an explicit `return 200` within the if/else block? I was trying to train myself not to do that! :) If that's what you mean, yes I would be happy to.

config/initializers/booleanizer.rb
```diff
@@ -0,0 +1,7 @@
+class String
+  def to_bool
```
jessieay, line 2: it might be easier to just require this value to be `'true'` or `'false'` than the monkeypatch `String`, but I am not super opposed

CONTRIBUTING.md
```diff
@@ -57,6 +57,35 @@
 
 Server tests are in rspec. Just run `rspec`.
 
+### Load Tests
+
+Load testing requires manually installing some gems.
```
jessieay, line 62: maybe link to `tourbus` docs somewhere?

config/initializers/booleanizer.rb
pkarman, line 2: I often forget that `FOO=1` and `FOO=true` and `FOO=yes` all look like "truth" but aren't.

Similar to this http://www.rubydoc.info/github/gradeawarrior/ruby-development-toolbox/String%3Ato_bool

app/controllers/api/v0/signatures_controller.rb
jessieay: argh sorry for lack of clarity, what I meant was I think you can do this:

``` ruby
def update_manifest(cdx_response, signature_request, manifest)
  if cdx_response.key?(:document_id)
    manifest.document_id = cdx_response[:document_id]
    manifest.activity_id = signature_request[:activity_id]
    manifest.save!
    200
  else
    422
   end
end
```

and it will be equal to what you have here

app/controllers/api/v0/signatures_controller.rb
pkarman: I prefer that too, actually. Lazy me.

config/initializers/booleanizer.rb
jessieay, line 2: i :ear: ya

CONTRIBUTING.md
pkarman, line 62: good idea, added in ab6793a

jessieay:
I am ready to merge this once CI passes.

PR #102, opened 2016-01-23
always treat API controller body as JSON, even for create (POST)

pkarman:


app/controllers/api/v0/manifests_controller.rb
```diff
@@ -12,30 +8,31 @@ def search
   end
 
   def validate
-    begin
-      manifest_content = JSON.parse(request.body.read)
+    manifest_content = read_body_as_json
+    unless performed?
```
jessieay, line 12: where does `performed?` come from in this context? I would guess it is related to an async process but we are not using those for API calls as far as I know so the naming is a bit confusing to me.

pkarman, line 12: `performed?` is just avoiding multi-render bug. http://apidock.com/rails/ActionController/Metal/performed%3F

Since we have consolidated the error handling, it's possible we have already rendered an error before we get to the success render.

spec/docs/swagger_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 include JsonSchemaSpecHelper
 
-describe 'Swagger docs', type: :apivore, order: :defined do
+describe 'Swagger docs', type: :apivore, order: :defined, elasticsearch: true do
```
jessieay, line 4: interesting, why do we need es for these specs?

app/controllers/concerns/manifest_params.rb
```diff
@@ -1,5 +1,12 @@
 module ManifestParams
+  def deep_reject!(hash, &block)
```
jessieay, line 8: could we write a test for this method? seems a bit complicated to be untested

app/controllers/api/v0/manifests_controller.rb
jessieay, line 12: oh interesting - where might we have already returned an error msg in this case? Perhaps writing a test to cover that case would clarify the need to call `performed?`

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -12,30 +8,31 @@ def search
...
-        tracking_number = manifest_params[:manifest_tracking_number]
-        render json: {
-          message: "Manifest #{tracking_number} submitted successfully.",
-        }, status: 201
-      else
-        render json: {
-          message: "Validation failed",
-          errors: @manifest.errors.full_messages.to_sentence
-        }, status: 422
+        if @manifest.save
```
jessieay, line 25: I know we aren't changing any logic here, but seeing this controller again I am wondering -- in what case would the manifest JSON be valid but the manifest would not save? if there is no such case, perhaps we don't need a second `if` clause for `manifest.save`

app/controllers/api/v0/manifests_controller.rb
```diff
@@ -47,23 +44,21 @@ def show
 
   def update
     manifest = find_manifest
+    patch = read_body_as_json
 
-    patch = JSON.parse(request.body.read)
-    patch_json = patch.to_json
+    unless performed?
+      patch_json = patch.to_json
```
jessieay, line 50: perhaps we could just do `patch_json = read_body_as_json.to_json` here?

app/controllers/manifests_controller.rb
```diff
@@ -26,4 +30,14 @@ def index
   def show
     @manifest = Manifest.find_by_uuid_or_tracking_number!(params[:id])
   end
+
+  private
+
+  def validate_manifest(content)
```
jessieay, line 36: looks like we have this same method in two places -- perhaps put in the shared concern?

app/controllers/api/v0/manifests_controller.rb
pkarman, line 12: if `read_body_as_json` fails to parse JSON, we get a 400 error. That is tested here:

https://github.com/18F/e-manifest/blob/master/spec/requests/api/v0/validate_manifest_spec.rb#L4

In that case, we must check immediately after `read_body_as_json` is called in order to avoid double render.

spec/docs/swagger_spec.rb
pkarman, line 4: there's a call to `/api/v0/manifests/search`

app/controllers/api/v0/manifests_controller.rb
pkarman, line 25: Right now, you are correct. Since the only validation in the model is redundant with what is done via the JSON schema, we should not see the case you're describing.

However, I don't think it's a bad idea for more error checking. The alternative would be calling `@manifest.save!` in which case something like a psql error should also surface (not just validation errors but anything that fails in the INSERT or UPDATE).

app/controllers/api/v0/manifests_controller.rb
pkarman, line 50: `read_body_as_json` may return nil if the body is invalid JSON.

app/controllers/manifests_controller.rb
pkarman, line 36: is it the same? I think in the API controller the validate_manifest method renders a JSON response and in the web controller validate_manifest sets the `@errors` only.

app/controllers/concerns/manifest_params.rb
pkarman, line 8: good idea. I will write a unit test for it.

The existing feature tests will fail without it though; the web form submits lots of empty fields.

app/controllers/manifests_controller.rb
jessieay, line 36: you are correct!

app/controllers/api/v0/manifests_controller.rb
jessieay, line 25: ok sounds good

app/controllers/concerns/manifest_params.rb
jessieay, line 8: noted. don't think we should get rid of the method, just want to see it under test so we / I don't break its functionality in the future

spec/docs/swagger_spec.rb
jessieay, line 4: got it!

jessieay:
Le me know when the `deep_reject!` spec is added. otherwise, looks good to go to me!

pkarman:
@jessieay test added in  63d6cf4

PR #103, opened 2016-01-25
Signing token date

pkarman:
This PR does 2 things:
- CDX signature token is stored server-side in Redis and referenced by a random uuid, which is what is passed on in our _token_ response. This is because CDX returns a single token for all e-manifest requests. Previously this was done with a session id but sessions in Rails are stored in cookies; using Redis lets us scale our web servers as well.
- Stores the signature datetime on the manifest itself. This makes explicit when the manifest was signed, which avoids the case where the manifest is updated after being signed and we can therefore no longer rely on the `updated_at` value.

This PR is branched off of #101 and should be merged after that PR is merged.

### DIFF
--- a/app/controllers/api/v0/tokens_controller.rb
+++ b/app/controllers/api/v0/tokens_controller.rb
@@ -19,6 +19,8 @@ class Api::V0::TokensController < ApiController
     end
   end
 
+  private
+
   def store_signature_token(cdx_token)
     user_token = SecureRandom.uuid
     redis = Redis.new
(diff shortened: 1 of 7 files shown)


app/controllers/api/v0/tokens_controller.rb
```diff
@@ -17,4 +18,11 @@ def create
       }.to_json, status: 401
     end
   end
+
+  def store_signature_token(cdx_token)
```
jessieay, line 24: should we write some tests around this new behavior?

jessieay:
Would be nice to have some tests around the redis token storage.

otherwise, I am :+1: on all of this.

pkarman:
redis test added in 1bfb491

pkarman:
@jessieay all :green_heart: again

jessieay:
@pkarman merged!

PR #104, opened 2016-01-25
Swagger doc fixes

pkarman:


jessieay:
This looks great! I just refreshed travis because that error looks unrelated. Can't wait until we can turn off these logs, makes test output so difficult to read!

PR #105, opened 2016-01-25
catch non-unique manifest tracking number on create

pkarman:


jessieay:
this code looks good. 

what about adding a uniqueness validation to manifest tracking # instead? That would be less controller code and would ensure we don't create manifests with duplicate manifest tracking numbers anywhere in our app.

pkarman:
@jessieay there is already a psql unique constraint for tracking number on the db, which is tested for and gets triggered correctly. What I am trying to do here is give a nicer error message for that particular use case, before the exception is triggered.

jessieay:
@pkarman what about an activerecord validation?

pkarman:
ah I see what you mean. Yes, that's better. Will do.

pkarman:
@jessieay validation on model now.

app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,33 @@
+module Searchable
```
jessieay, line 1: nice! I have also been wanting to add this concern. LOVE IT

app/models/manifest.rb
```diff
@@ -1,7 +1,9 @@
 require 'date'
 
 class Manifest < ActiveRecord::Base
-  validates :tracking_number, presence: true
+  include Searchable
+
+  validate :tracking_number, :validate_tracking_number_unique
```
jessieay, line 6: how is `validate_tracking_number_unique` giving a different result than doing `validates :tracking_number, presence: true, uniqueness: true` ? (built in AR validations)

pkarman, line 6: I tried `uniqueness: true` at first too, but since `tracking_number` isn't a proper database column, AR doesn't know how to generate SQL for it.

jessieay, line 6: got it, that makes sense.

PR #106, opened 2016-01-25
catch case where manifest.generator_mailing_address is nil

pkarman:


spec/models/manifest_spec.rb
```diff
@@ -97,4 +97,22 @@
...
+
+        expect(manifest.generator_mailing_address).to eq({ 'zip_code' => '12345' })
+      end
+    end
+
+    context 'generator_mailing_address is absent' do
+      it 'returns empty hash' do
+        manifest = build(:manifest)
+
+        expect(manifest.generator_mailing_address).to eq({})
```
jessieay, line 114: this brings up a larger question of whether we want to return keys for fields that have no data 

eg: return `{ 'zip_code' => '' }` here

I think it can be helpful to return empty values for all fields, but not something we have to tackle right now.

PR #107, opened 2016-01-25
Update signature request specs to use VCR

jessieay:
- Update web signature controller to actually sign, too
- Add JSON schema for validating sign requests


public/schemas/post-signature.json
```diff
@@ -0,0 +1,41 @@
...
+      "type": "string"
+    },
+    "token": {
+      "description": "security token from CDX API",
+      "type": "string"
+    },
+    "user_id": {
+      "description": "user_id used to access CDX API",
+      "type": "string"
+    }
```
jessieay: I need to add `manifest_content` to this lsit

app/controllers/api/v0/signatures_controller.rb
```diff
@@ -19,29 +13,14 @@ def create
 
   def prep_signature_request(manifest)
     signature_request = read_body_as_json(symbolize_names: true)
-    signature_request[:manifest_content] = manifest.content.to_json
-    if signature_request[:token]
-      signature_request[:token] = lookup_signature_token(signature_request[:token])
-    else
-      render json: { message: "Missing signature token" }, status: 403
```
jessieay, line 26: I removed this because I think we need to deal with this situation in a more detailed way. Right now, a missing `token` arg returns this message. But if the token is wrong, we return a 422 and different error message. Going to work on improving the error message for both cases.

jessieay, line 26: ok, now a missing token param will be caught by JSON schema validator

app/services/base_validator.rb
```diff
@@ -0,0 +1,33 @@
+class BaseValidator
```
jessieay, line 1: maybe we should put these validator classes into `app/validators` -- thoughts?

app/services/manifest_signer.rb
```diff
@@ -0,0 +1,46 @@
...
+    @args = args
+  end
+
+  def perform
+    cdx_start = Time.current
+
+    cdx_response = CDX::Manifest.new(parsed_args).sign
+
+    cdx_stop = Time.current
+    Rails.logger.debug(ANSI.blue{ "  CDX signature time: #{sprintf('%#g', (cdx_stop - cdx_start))} seconds" })
```
jessieay, line 14: should we add this kind of logging for everything or remove it from here? seems weird to have it in one CDX API caller class but not the other

app/services/base_validator.rb
pkarman, line 1: Doesn't matter to me.

app/services/manifest_signer.rb
pkarman, line 14: I think it's good to have it somewhere, outside the CDX classes themselves. I'm interested in the total run time of the CDX action, from the perspective of the caller.

pkarman:
+1 on the use of schema validators and the resulting refactor.

If you want to rename `app/services` to `app/validators` I'm fine with that. I will merge whenever you give the :+1:

pkarman:
@jessieay I made some changes to quiet the CDX logger during tests. Hurrah!

jessieay:
Happy to keep in services for now. Perhaps move later. Merge away!

PR #110, opened 2016-01-26
use ruby Array.join method instead of helper

pkarman:


jeremiak:
Great, thanks @pkarman - looks good to me

PR #113, opened 2016-01-26
Search pagination

pkarman:
Trello: https://trello.com/c/8DGCW7wA

### DIFF
--- a/app/views/manifests/index.html.erb
+++ b/app/views/manifests/index.html.erb
@@ -1,6 +1,10 @@
 <div class="usa-width-two-thirds">
 
-  <h3>Public manifests</h3>
+  <h3>Manifests</h3>
+
+  <div class="stats">
+    Currently viewing <%= @manifests.size %> of <%= number_with_delimiter @es_response.results.total %> results.
+  </div>
 
   <table>
     <thead>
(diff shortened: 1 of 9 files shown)


app/views/manifests/index.html.erb
```diff
@@ -1,6 +1,10 @@
 <div class="usa-width-two-thirds">
 
-  <h3>Public manifests</h3>
+  <h3>Manifests</h3>
+
+  <div class="stats">
+    Currently viewing <%= @stats[:from] %>-<%= @stats[:to] %> of <%= number_with_delimiter @stats[:total] %> results.
```
jessieay, line 6: not sure what the standard # of records for pagination is or how it is being set -- would it be possible to write a feature level test for this pagination? Ideally, we could set records per page to "1" and test that when we create 2 records that match a seacrch, only one of them is on the page.

pkarman, line 6: agreed about need for tests. added in 983f28d

spec/requests/manifests_search_spec.rb
```diff
@@ -8,5 +8,11 @@
       get "/manifests?aq[content.generator.name]=company"
       expect(response.body).to include(manifest.tracking_number.to_s)
     end
+
+    it 'paginates' do
+      10.times { create(:manifest, :indexed) }
```
jessieay: what about creating 5 records? that would test that pagination works just as well as creating 10 records, wouldn't it?

pkarman:
@jessieay test updated, all :green_heart:

jessieay:
@billHolochwostCGI did you mean to close this PR?

pkarman:
codeclimate appears stuck. when I click the link, no errs found.

PR #116, opened 2016-01-26
Fix params in uploads controller

jessieay:
- Was not previously actually uploading files
- https://trello.com/c/5Z6AT90F

### DIFF
--- a/app/controllers/manifest_uploads_controller.rb
+++ b/app/controllers/manifest_uploads_controller.rb
@@ -22,14 +22,18 @@ class ManifestUploadsController < ApplicationController
       manifest.content[:uploaded_file] = image_details
       manifest
     else
-      Manifest.new(content: manifest_upload_params)
+      Manifest.new(content: parsed_upload_params)
     end
+  end
+
+  def parsed_upload_params
+    manifest_upload_params.merge(uploaded_file: image_details)
   end
 
   def manifest_upload_params
     params.require(:manifest).permit(
       generator: [:manifest_tracking_number],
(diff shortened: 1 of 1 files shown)


pkarman:
was there a test for this? curious that if there was it was passing before?

jessieay:
there is a feature spec for file upload that was passing because the strong params were defined incorrectly, which means that the manifest was being created with just a tracking number and null file uploaded in the manifest content.

I tried writing a request spec to test for what is happening at a lower level, but was struggling with getting the file go go through the params correctly. In the web controller, it comes through as an instance of http://api.rubyonrails.org/classes/ActionDispatch/Http/UploadedFile.html, but when I created an UploadedFile and sent it through the params in a request spec, it was turned into a string. 

After struggling with that for a little, I jumped ship because it was taking more time than seemed worth it.

One thing we could do would be add a validation via JSON schema that if the `uploaded_file` key is present, it contains the `file_name` `content` and `content_type` sub keys/values. What do you think?

jessieay:
also just FYI, this is the spec: https://github.com/18F/e-manifest/blob/master/spec/features/upload_scan_spec.rb

once strong params were defined appropriately, they failed until the params were parsed correctly. would still be nice to have a test that confirms file fields are present, though

pkarman:
I agree that validating the JSON would be ideal.

pkarman:
I'm ok to merge this as-is and add a new card for the JSON spec.

jessieay:
@pkarman sounds good to me!

PR #117, opened 2016-01-26
Allow download of scanned images

jessieay:
With scanned image:

screen shot 2016-01-26 at 2 27 52 pm

Without scanned image:

screen shot 2016-01-26 at 2 27 40 pm

PR #118, opened 2016-01-26
Add helpful error messages for web when invalid manifest

jessieay:
- Formerly, was showing json_schema errors first, which were hard to
  understand

Before:

screen shot 2016-01-26 at 3 00 57 pm

After:

screen shot 2016-01-26 at 3 01 06 pm

### DIFF
--- a/app/controllers/manifests_controller.rb
+++ b/app/controllers/manifests_controller.rb
@@ -5,17 +5,16 @@ class ManifestsController < ApplicationController
   end
 
   def create
-    if validate_manifest(manifest_params)
-      @manifest = Manifest.new(content: manifest_params)
+    @manifest = Manifest.new(content: manifest_params)
 
-      if @manifest.save!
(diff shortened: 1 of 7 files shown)


pkarman:
a visual nit: I could not see the error message in the "after" screenshot at first because it is in the blue header banner. Could we stick it in the body, framed in some kind of border/color?

jessieay:
Do you think we would move the location of all flashes? Would be annoying
to move one type and not all.

pkarman:
Yes I would move them all.

jessieay:
@pkarman just realized that this is not the branch with the updated flash styles. Here is a screenshot of a flash from https://github.com/18F/e-manifest/pull/117

screen shot 2016-01-27 at 10 08 43 am

you still think we should move it?

pkarman:
ah that's much better. no need to move it.

PR #120, opened 2016-01-27
WIP [Manifest form parity]

jessieay:


jessieay:
@pkarman I am opening this now as a WIP because:
- I believe this will be the simplest lib for having form parity with the old form for array fields that could be of any length (transporters and vendors) https://github.com/nathanvda/cocoon
- That lib requires `accepts_nested_attributes`
-  after struggling with allowing nested attributes for a non-activerecord object for a bit, I got stuck (see: https://github.com/joker1007/activemodel-associations/issues/8)
- I am unsure if this is a road we want to go down. We could easily have the API accept and show the same JSON it currently shows if we have a separate table for `vendors` rather than storing it in `content`, but this is a move away from the current strategy of having _almost_ all fields stored in `Manifest#content` (which isn't a strategy I am particularly attached to, but it is the current strategy)

Let me know what you think?

jeremiak:
@jessieay I fixed the spacing a tiny bit on that download callout box and committed to this branch

pkarman:
I am very reluctant to start adding additional manifest-related models at this point. That introduces a hybrid data model (some stuff in JSON blob, some stuff in different table(s)) that will be hard to rollback.

Is the dilemma that adding dynamic form fields to the UI is difficult w/o using javascript? We already load jquery. Can we just add the form fields dynamically with a little js?

Changing the data model architecture in order to solve a UI problem feels less than ideal.

jessieay:
@pkarman ok that's what I thought. I agree with you that changing the data model could become problematic. Still not sure I think the current model is what we should stick with for all of time, but given that we have ~3 days left on the proj, now is probably not the time to start moving things over.

brandocalrissian:
Chiming in agreement about avoiding hybridizing the manifest data model unless absolutely necessary. Maintenance or migration to a different data store becomes confusing and difficult.

jessieay:
One benefit of moving some attributes of a manifest's content into a separate database column or table is that we would have to fight Rails a lot less. 

In the land of Rails convention, data is most easily accessed via individual database columns or associated database tables. Dealing with this manifest form was much more difficult because all of the content lives in a single json column. 

As long as this is a Ruby/Rails project, it will be a big development time saver to use Rails the Rails way. That doesn't mean moving everything over right away, but I do think that moving manifest content into separate database columns or tables should be seriously considered.

This would not affect the way that the data is Serialized, we would just need to update the Serializer to display the data based on database columns rather than a single json `content` column.

bbingCGI:
+1 @jessieay I'm not a rails guru.  But I believe one of the best perks when using rails is its ability to rapidly ORM your data layer all the way up to your presentation layer without much code/effort.  My rails gurus tend to agree.  How much effort is this versus reward?  I believe it's little.

On the flip side of that coin, I would be interested to see ElasticSearch and this field handle ad hoc queries across millions of records.  I'm curious to see its potential.  Is it wise to verify its abilities here?  No.  I'd rather see this in a performance test ahead of schedule.

PR #121, opened 2016-01-27
Update _manifest.json

scottdchristian:
Since this is an example, I moved the one mandatory field (manifest_tracking_number) to the top so that people testing the app can easily find it and edit it.  I thought about changing the value in the example for "manifest_tracking_number" to a note that would not validate, but decided not to. here is an example of that thought:

```
  "manifest_tracking_number" : "ENTER A STRING OF 9 Digits and 3 Character, for example 123456789ABC"
```

Would this be useful or should we leave it as an example.

### DIFF
--- a/app/views/examples/_manifest.json
+++ b/app/views/examples/_manifest.json
@@ -1,5 +1,6 @@
 {
    "generator" : {
+      "manifest_tracking_number" : "987654321ABC",
       "us_epa_id_number" : "ABC712334321",
       "emergency_response_phone" : "888-999-8888",
       "signatory" : {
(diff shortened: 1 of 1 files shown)


pkarman:
@scottdchristian I would suggest leaving the example value as-is. The goal of the examples is that they would validate with a simple copy/paste, which is how they are tested as well.

scottdchristian:
Thanks that was what I thought the answer would be thus what I pushed, but I figured could not hurt to ask, since the idea came to mind. 

 On a side note (not important) any idea why I could edit this file in the GitHub webpage and not the swagger.json or API docs?

pkarman:
I'm not familiar enough with the github editor to answer.

scottdchristian:
The file as proposed is a minor change to move the manifest_tracking_number.  As user of the example I would prefer to see the item I need to change on line 3 rather than scanning down 20 odd lines to find the item I need change.

My other item was a question, which was not included in the file.

PR #123, opened 2016-01-27
Update post-manifest.json

scottdchristian:
Added in the density field, for the biennial report


pkarman:
not sure why travis is puking about missing `secret_key_base` when other PRs are not.

Going to merge this since it is so small.

scottdchristian:
Thank you, I assume Travis issue is probably because I did the pull request directly in GitHub rather than from my machine.

I'll bet my other PR has the same issue.

PR #124, opened 2016-01-27
Add native users + authentication

pkarman:
Cards: 
- https://trello.com/c/sj4Vk0hc
- https://trello.com/c/r8BXuqJU

Allows for users to login/out using CDX credentials, applies some basic authorization sanity for creating/updating/searching/viewing manifests.


spec/requests/api/v0/post_tokens_spec.rb
```diff
@@ -2,35 +2,27 @@
 
 describe 'POST /api/v0/tokens' do
   it 'returns a 401 and no token when auth fails' do
-    VCR.use_cassette('auth_failure') do
```
jessieay, line 5: we can remove this cassette if we are no longer using it. do you think we _should_ be using it somewhere if not here? I am not sure what the best line is between mocking CDX and using VCR

spec/requests/api/v0/post_manifest_spec.rb
```diff
@@ -1,6 +1,9 @@
 require 'rails_helper'
 
 describe 'POST manifests' do
+  before(:each) do
+    @current_session = mock_authenticated_session
```
jessieay: inline this? I know it's more code, but I like seeing the setup required for each spec!

jessieay: also, do we need an authed session in order to create a manifest? was not sure if that was required or not

spec/models/user_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  describe 'cdx_user_id' do
+    it 'is always required' do
+      user = create(:user)
+      expect(user.cdx_user_id).to_not be_nil    
+    end
+
+    it 'is unique' do
+      user = create(:user)
+      expect {
+        create(:user, cdx_user_id: user.cdx_user_id)
```
jessieay, line 17: I think for this you could just use shoulda for this: `it { should validate_uniqueness_of(:cdx_user_id) }`

spec/models/user_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+describe User do
+  describe 'Associations' do
+    it { should have_many(:manifests) }
+  end
+
+  describe 'cdx_user_id' do
+    it 'is always required' do
+      user = create(:user)
+      expect(user.cdx_user_id).to_not be_nil
```
jessieay, line 11: could also use shoulda for this, I think: `it { should validate_presence_of(:cdx_user_id) }`

spec/models/manifest_spec.rb
```diff
@@ -148,4 +148,19 @@
       end
     end
   end
+
+  describe '#is_public?' do
```
jessieay, line 152: nice! I think `public?` would be as clear as `is_public?` but either works

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  controller do
+    def index
+      render inline: 'hello world'
+    end
+  end
+
+  describe 'user session' do
+    it 'extends life of session each time user makes a request' do
+      initial_session_time = @current_session.updated_at
+      sleep 1
```
jessieay, line 17: can we use timecop rather than sleep?

db/seeds.rb
```diff
@@ -5,3 +5,5 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+
+User.create(id: 0, cdx_user_id: 'unknown_user')
```
jessieay: should we have a note in the readme about this? (not obvious to be why we need it / how it is used)

spec/requests/api/v0/post_tokens_spec.rb
pkarman, line 5: after spending some time with this, my suggestion would be to use the VCR approach with the unit tests in `spec/models/cdx` and the mock approach with everything else. That gives the most granular control and reduces the maintenance burden on us to keep VCR scripts in sync with reality.

spec/requests/api/v0/post_manifest_spec.rb
pkarman: We need a user to associate with each manifest, so yes, authentication is required.

I will inline; the fact that I didn't indicates I didn't add enough tests for authz.

spec/models/user_spec.rb
pkarman, line 11: if that worked, it would be very cool.

here's what I tried and failed:

``` diff
diff --git a/spec/models/user_spec.rb b/spec/models/user_spec.rb
index 4ae36a3..0e6b4ac 100644
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -3,6 +3,8 @@ require 'rails_helper'
 describe User do
   describe 'Associations' do
     it { should have_many(:manifests) }
+    it { should validate_uniqueness_of(:cdx_user_id) }
+    it { should validate_presence_of(:cdx_user_id) }
   end

   describe 'cdx_user_id' do
```

```
User
  Associations
    should validate that :cdx_user_id is case-sensitively unique (FAILED - 1)
    example at ./spec/models/user_spec.rb:7 (FAILED - 2)
    should have many manifests
  cdx_user_id
    is always required
    is unique
    #find_or_create

Failures:

  1) User Associations should validate that :cdx_user_id is case-sensitively unique
     Failure/Error: it { should validate_uniqueness_of(:cdx_user_id) }
       User did not properly validate that :cdx_user_id is case-sensitively
       unique.
         The record you provided could not be created, as it failed with the
         following validation errors:

         * cdx_user_id: ["can't be blank"]
     # ./spec/models/user_spec.rb:6:in `block (3 levels) in <top (required)>'

  2) User Associations 
     Failure/Error: it { should validate_presence_of(:cdx_user_id) }
     NoMethodError:
       undefined method `validate_presence_of' for #<RSpec::ExampleGroups::User::Associations:0x007ff6ca35f138>
     # ./spec/models/user_spec.rb:7:in `block (3 levels) in <top (required)>'

Finished in 0.17316 seconds (files took 3.94 seconds to load)
6 examples, 2 failures
```

spec/models/manifest_spec.rb
pkarman, line 152: the English student in me always likes the explicit verb but I see that Ruby conventions tend toward brevity.

http://programmers.stackexchange.com/questions/232213/should-i-always-use-is-as-prefix-for-boolean-variables

I find this more readable, personally:

```
if thing.is_foobar?
```

than this:

```
if thing.foobar?
```

because the former reads like an actual sentence, and clarifies its being-ness rather than its acting-ness. The `?` indicates that the return value is a boolean, but does not necessarily imply that the method is idempotent.

ok, spiel over. if you don't care much, let's leave as is.

spec/controllers/application_controller_spec.rb
pkarman, line 17: that was my first instinct too.

alas, the feature depends on the Redis server, which is completely independent of Timecop and of Ruby.

db/seeds.rb
pkarman: I needed it at one point in my dev but will remove.

spec/requests/api/v0/post_manifest_spec.rb
pkarman: fixed in 945e714

spec/models/user_spec.rb
jessieay, line 11: ah yes there is a weird bug in shoulda https://github.com/thoughtbot/shoulda-matchers/blob/585f651460026045f590ea27e14f6c5f09e8e6c4/lib/shoulda/matchers/active_model/validate_uniqueness_of_matcher.rb#L28

related to combo presence and uniqueness validation,s should be able to do:

```
subject { build(:user) }
it { should validate_uniqueness_of(:cdx_user_id) }
```

not sure what the other error is for the presence validation. if this is too much of a pain we can just skip it for now

spec/requests/api/v0/post_tokens_spec.rb
jessieay, line 5: sounds good :+1:  - let's remove these unused scripts, then!

app/services/user_session.rb
```diff
@@ -0,0 +1,124 @@
+class UserSession
+  attr_reader :session, :token
+
+  TTL = 4 * 60 * 60
+  @@ttl = TTL
+
+  @@namespace = "user_session"
```
jessieay, line 7: Hmmm I do not see class vars used to much. Mind making this a class method instead? I understand their use for caching below but might be easier to read if this is just a getter method

PR #125, opened 2016-01-27
Small schema changes

jessieay:
- Update typo on container type enum
- Use date type for date fields rather than 3 separate fields
- Remove unnecessary boolean fields
- Got verbal sign off on these changes from Scott


pkarman:
big :+1:

PR #126, opened 2016-01-28
Add fields to Manifest form

jessieay:
- Going for parity with Sinatra app
- Still need to add ability to add multiple transporters / manifest items
- Also need to parse params correctly in controller, but this is close

### DIFF
--- a/app/controllers/manifests_controller.rb
+++ b/app/controllers/manifests_controller.rb
@@ -8,12 +8,10 @@ class ManifestsController < ApplicationController
     @manifest = Manifest.new(content: manifest_params)
 
     if @manifest.save && validate_manifest(manifest_params)
-      @manifest.reload
-      tracking_number = @manifest.tracking_number
-      flash[:notice] = "Manifest #{tracking_number} submitted successfully."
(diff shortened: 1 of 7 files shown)


pkarman:
looks great. waiting for :green_heart: and will merge.

jessieay:
@pkarman :green_heart: now -- darn shard failures!

app/controllers/manifests_controller.rb
```diff
@@ -8,12 +8,10 @@ def create
     @manifest = Manifest.new(content: manifest_params)
 
     if @manifest.save && validate_manifest(manifest_params)
```
pkarman, line 10: this introduces a race condition. i.e., the manifest could be saved w/o being validated against the schema.

Look at https://github.com/18F/e-manifest/commit/aa7d449e2d456bf6c4ca0224f5ba17196d0701ec

jessieay, line 10: ok will update

pkarman, line 10: funny that we were both solving for this, in a similar way. :)

PR #132, opened 2016-01-29
zero-downtime-push does not preserve env vars from *-venerable rename

pkarman:
zero-downtime renames the live instance, then creates a new one. But it fails to copy over the env from the existing live instance. This PR creates a temp manifest and deploys using it, so env is preserved.

### DIFF
--- a/script/deploy
+++ b/script/deploy
@@ -14,4 +14,9 @@ if ! hash autopilot 2>/dev/null; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


brandocalrissian:
When I push this branch to dev I still get the "Unable to authenticate user - The password is invalid. Could not perform central authentication for user <> in domain <default>" error when trying to log in.

brandocalrissian:
Works for me now.

PR #134, opened 2016-01-29
Add discrepancy fields

jessieay:
- Include manifest tracking #, which is not on prod
- Next up: test it! And add params parsing, which we will need

https://trello.com/c/JnYKDncA

### DIFF
--- a/public/schemas/post-manifest.json
+++ b/public/schemas/post-manifest.json
@@ -378,7 +378,7 @@
               "enum":["no_rejection","partial_rejection","full_rejection"]
             },
             "description": {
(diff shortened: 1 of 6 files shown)


public/schemas/post-manifest.json
```diff
@@ -378,7 +378,7 @@
               "enum":["no_rejection","partial_rejection","full_rejection"]
             },
             "description": {
-              "description": "(14) Description of discrepancy",
+              "description": "(18) Description of discrepancy",
```
jessieay, line 381: this is line with what I am seeing on prod and the rest of the fields in this section. 90% sure it was a typo here.

PR #135, opened 2016-01-29
Add toggle for international fields

jessieay:
Merge after https://github.com/18F/e-manifest/pull/134

### DIFF
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -50,6 +50,9 @@ class Manifest < ActiveRecord::Base
     content_field('designated_facility.name')
   end
 
+  def international_shipment
+  end
+
   def designated_facility_signed_date
     if content_field('designated_facility')
       certification = content_field('designated_facility.certification')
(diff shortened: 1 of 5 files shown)


app/models/manifest.rb
```diff
@@ -50,6 +50,9 @@ def designated_facility_name
     content_field('designated_facility.name')
   end
 
+  def international_shipment
+  end
```
pkarman, line 54: what's with this no-op?

jessieay, line 54: when the form renders on a failed submit, Rails tries to autofill form fields based on database attributes / model methods. since we have a boolean for `international_shipment`, it gets a an `undefined method` error on render without this.

there is a larger issue with this form, which is that right now _some_ field values are retained on render, but not all because of our unusual database setup. going to create a card for this because it would be possible to retain all values for the `content` on a `render :new` if we defined methods for each one.

jessieay, line 54: card created: https://trello.com/c/ckMYsL79

PR #142, opened 2016-02-05
[WIP] CDX Roles sync

pkarman:
Trello: https://trello.com/c/BBTE9ToE and https://trello.com/c/ep2wQ2cl

This PR adds support for fetching and storing CDX Organizations and Roles. There is also a rationale document.

The PR is getting big so I wanted to get it up even though there are still some TODOs:
- ~~VCR tests for new CDX::\* classes~~ (#145)
- background worker to run `UserProfileSyncer` every time a user logs in
- add Pundit to start fleshing authz rules architecture

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,6 +122,8 @@ The `CDX_LOG` and `CDX_COLOR` environment variables control the verbosity and st
 model logging. Currently the `CDX_LOG` variable must be set to `true` when running the rspec test suite,
 and therefor is explicitly set in the `.env.test` file.
 
+See also the [Authentication and Authorization architecture documentation](doc/authorization.md).
+
 ## Public domain
 
 This project is in the public domain within the United States, and
--- a/app/models/cdx/client.rb
+++ b/app/models/cdx/client.rb
@@ -4,6 +4,7 @@ class CDX::Client
   CDX_BASE_URL = ENV['CDX_BASE_URL'] || "https://devngn.epacdxnode.net"
   CDX_SIGN_SERVICE_URL = ENV['CDX_SIGN_SERVICE_URL'] || "#{CDX_BASE_URL}/cdx-register/services/RegisterSignService?wsdl"
   CDX_AUTH_SERVICE_URL = ENV['CDX_AUTH_SERVICE_URL'] || "#{CDX_BASE_URL}/cdx-register/services/RegisterAuthService?wsdl"
+  CDX_AUTHZ_SERVICE_URL = ENV['CDX_AUTHZ_SERVICE_URL'] || "#{CDX_BASE_URL}/cdx-register/services/RegisterService?wsdl"
 
   def initialize(opts={})
     @savon = Savon.client(default_opts.merge(opts))
@@ -15,7 +16,7 @@ class CDX::Client
       pretty_print_xml: true,
       log: (ENV['CDX_LOG'].present? ? ENV['CDX_LOG'].to_bool : false),
       soap_version: 2,
-      convert_request_keys_to: :none,
+      convert_request_keys_to: :lower_camelcase
     }
   end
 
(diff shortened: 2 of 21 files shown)


brandocalrissian:
I believe I'm following all the code in the PR after going over it a few times. I can predict how elasticsearch will index the new fields to drive role- and organization-based data access authorization and see how it fetches the organization and role data from the CDX service.

I'm still a bit wary of "on delete cascade" like a bad flashback. We can certainly delete records from the user_org_roles mapping table as they go out of sync, but will we ever delete anything out of the organizations, roles, and users tables? If not, do we need to worry about cascading deletes to the user_org_roles table?

pkarman:
organizations, roles, and user_org_roles are all cached data from CDX. There's nothing special or unique about that data in the e-manifest system. We cache it in psql just to make it easier to index and apply authz rules.

We also want it to get cleaned up automatically if we need to "nuke" the cache.

pkarman:
i.e. yes, I expect we will delete orgs, roles and even users, eventually. when that happens we do not want orphaned m2m user_org_role records mucking up the db.

brandocalrissian:
Agreed about CDX being the system of record and am on the same page about caching the data in the tables. I'm fine with delete cascade as long as we know why we're doing it. PR merged.

Thank you!

bbingCGI:
I'm concerned about this implementation.  I assumed that the CDX roles would be pulled and applied against a role matrix for a permission set.  The permission set would be used thereafter.  For all things involving security, checks should continue to be made from the source regardless of performance.  

I'm not sure how much of a performance gain you get from this.  If calls to the API are an issue, then the API needs to be updated.  e-Manifest shouldn't reduce its security context.  For CDX and SCS, they have no performance issues; and CDX is hit hard .. as are all of the APIs they use such as CROMERR.  The intent of Shared CROMERR Services is to be able to maintain security when hit from hundreds of disparate services.  This is the case for CDX and all of the dataflows linked to it.

Here is one scenario I'm concerned with:
1)  Hacker goes into the system with a user’s credentials (because they used bad password practices or the like).
2)  CDX deactivates the user immediately.  User cannot access CDX now.
3)  e-Manifest continues to allow this user to do whatever until it checks again.

Am I overthinking this?  Sorry I'm late to respond.

pkarman:
@benjaminBinghamCGI 

Thanks for the detailed comments.

The point of caching the CDX roles+orgs is much less about performance and more about the ability to apply authorization checks to the Elasticsearch index search results. The roles+orgs would be re-synced every time the user authenticates against CDX (that piece is listed in the TODOs on this PR).

e-Manifest requires users to re-authenticate to sign manifests, so given your example scenario, if the user were deactivated at CDX, manifests could not be signed and posted to CDX.

Are you suggesting that every HTTP request to e-Manifest should re-authenticate against CDX? To avoid storing passwords within e-Manifest, that would mean the user would need to provide user/pass for every request.

Right now "signing in" to e-Manifest does a one-time credentialed authentication against the CDX, and initiates a session within the e-Manifest application. That session is expired after 4 hours.

bbingCGI:
Hmm.  I would prefer a check for every request.  How's the motto go? Your strongest security is your weakest link?  Having said that, it's still unreasonable to do all the work every time...

If the result from the CDX role request was cached and -- upon every X requests -- we polled the CDX roles again and compared results (which should be possible and very fast).  If it isn't 1-for-1, then update the role matrix?

For now, a reduced expiration would suffice if someone had a concern.  The more I think about this, the more this is simply a risk to put on our radar.

I'll bring this up to the others.  See if maybe even the API can provide a method for user authorization health checks.

pkarman:
The attack vector here seems very narrow.

If credentials are compromised, that might go unnoticed by CDX or e-Manifest administrators for an indefinite amount of time (e.g. days or weeks or even months). As soon as the compromise is discovered, and the account is deactivated/locked at CDX, then the race condition you've described takes effect for up to 4 hours. 

I.e., the villain might act for many days on either system, and then a few hours more on the e-Manifest system alone. That seems like a pretty small window of opportunity to search/update/create -- all actions which do not cross over into the system of record (CDX).

We can certainly set up sync checks at whatever interval seems reasonable. However, the org+role sync really doesn't provide the needed security check; it's the active/inactive status of the User itself that matters. In the scenario you've described, it would be enough to invalidate the current session at that point; the db sync is orthogonal to the authentication issue.

PR #143, opened 2016-02-08
Typo fixes + a couple of minor edits

brandocalrissian:
Saw a couple of typos in the text. Fixed them and gave the rest of the content a _very_ cursory edit pass.

Added some URLs for references-- when we split the content between pages, not every page wound up with a link to a previously referenced URL.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,12 +23,12 @@
 
 ### Development practices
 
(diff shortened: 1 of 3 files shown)


pkarman:
I believe `CDX_LOG` is no longer part of the `.env.test` file but we can fix that later.

PR #148, opened 2016-02-09
add Authorization header to swagger ui

pkarman:
Swagger UI needs to have Authorization header added to all HTTP requests because Rails session is not available in `/api` endpoints.


pkarman:
test failure should be fixed in #149

PR #150, opened 2016-02-09
CDX API has inconsistent case conventions

pkarman:


pkarman:
test failure should be fixed by #149

PR #166, opened 2016-02-24
Update index.html.md

scottdchristian:
Per Trello https://trello.com/c/ootI9GKv making this change.

### DIFF
--- a/app/views/api-docs/examples/index.html.md
+++ b/app/views/api-docs/examples/index.html.md
@@ -9,7 +9,7 @@ The production base URL is **https://e-manifest.18f.gov/api/v0**
 Notation note: path variables are expressed as `:variable_name`. The colon is not part of the URI.
 
 On this page you may learn about how to:
-
+* [Get an API Key] (#get-an-api-key)
 * [Validate a manifest](#validate-a-manifest)
 * [Generate an Authentication Token](#generate-an-authorization-token)
 * [Submit a manifest](#submit-a-manifest)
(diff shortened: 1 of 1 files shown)


app/views/api-docs/examples/index.html.md
```diff
@@ -20,6 +20,10 @@ On this page you may learn about how to:
 * [Sign a manifest](#sign-a-manifest)
 * [Management method codes](#management-method-codes)
 
+## Get an API Key
+If you are a developer from a TSDF and you'd like to request an API key so you can use the e-Manifest APIs, please check your organization's policies on implementing e-manifest.  Some companies may use unique API keys for each facility, others may use one API key for an entire company.  If you have questions please email emanifest@epa.gov (mailto:"emanifest@epa.gov).
```
pkarman: Did you want the `mailto` to be a hotlink? I think it will do that automatically without the explicit `mailto`.

pkarman:
Do we need to document (even for ourselves) how api.data.gov is related to e-manifest.18f.gov/api/ ? All the docs reference e-manifest.18f.gov and there is no code that checks for API keys.

app/views/api-docs/examples/index.html.md
scottdchristian: sweet thanks, should I just remove the parenthetical?

pkarman: yes

scottdchristian:
Yes we should where are the other docs?  I'll create a card in Trello regarding checking API keys

PR #171, opened 2016-03-03
Pundit roles

pkarman:
Trello: https://trello.com/c/LGkCxslV/333-only-allow-users-with-the-proper-cdx-roles-to-sign-a-manifest

This PR implements a few things:
- the `pundit` gem provides an easy way to define authorization rules per role. See https://github.com/elabs/pundit
- the ManifestPolicy ensures that only someone with a `TSDF Certifier` role in the org that "owns" the manifest can sign that manifest


pkarman:
@brandocalrissian this is ready for your review

PR #177, opened 2016-03-16
New API endpoint for submitting the manifest for CROMERR signing.

brandocalrissian:
Does not sign. Only TSDF users are allowed to use the API (same as signing rules).

Could use some more tests.

Does not include the bits that talk to the CDX services to perform the actual submission-- that's a different commit-- it takes the data to the front door of that request process.

Re-added the emacs autosave files to the .gitignore lost in the rails translation.

This is for https://trello.com/c/jHHqRC18.


brandocalrissian:
Code climate warnings are fair enough as it became clear it would be odd to support a way of performing this that was too far off of the signing path. Fixing...

brandocalrissian:
Pulling up or extracting the two similarity issues left in codeclimate seems artificial to me, especially as the signing support is going away.

I'm open to suggestion if there's a better way.

Artificial problem? I'll make an artificial solution. Once again, the code's going away in a later commit...

brandocalrissian:
ok, it replaced one artificial "you have the same call structure with different implementations" warning with another. I really don't care about this one.

I'm not going to extract the code to make it more meta and harder to follow.

app/controllers/api/v0/cromerr_submit_transactions_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Api::V0::CromerrSubmitTransactionsController < ApiController
```
pkarman: This controller name feels too long to me.

How about `Api::V0::SubmissionsController` instead?

app/controllers/api/v0/cromerr_submit_transactions_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  def create
+    submit_request = read_body_as_json(symbolize_names: true)
```
pkarman: `submit` is a verb. Maybe `submission` instead?

app/services/manifest_token_json_converter.rb
```diff
@@ -0,0 +1,22 @@
+class ManifestTokenJsonConverter
```
pkarman, line 1: This isn't really a service. Why did you feel it needed to be factored out of the controller? It appears to just mangle incoming params for use by an actual service.

app/services/manifest_submitter.rb
```diff
@@ -0,0 +1,26 @@
+class ManifestSubmitter
+  include Pundit
```
pkarman: Is `Pundit` needed here?

app/services/manifest_submitter.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def initialize(args)
+    @args = args
+  end
+
+  def perform
+    output_stream = StreamLogger.new(Rails.logger)
+
+    manifest = Manifest.find_by_uuid_or_tracking_number!(args[:manifest_id])
+    authorize manifest, :can_submit?
```
pkarman: ah I see.

I would suggest putting the authorization check _outside_ the service. Strong barriers at the castle walls mean the service can just do its one thing.

public/api-documentation/swagger.json
```diff
@@ -262,6 +262,46 @@
         }
       }
     },
+    "/manifests/{id}/cromerr_submit_transaction": {
```
pkarman: we generally try and keep verbs out of the URL. The verb is the HTTP method. The URL is the noun.

public/api-documentation/swagger.json
```diff
@@ -401,6 +441,32 @@
         }
       }
     },
+    "SubmitSignature": {
```
pkarman: `SignatureSubmission`?

public/api-documentation/swagger.json
```diff
@@ -401,6 +441,32 @@
...
+        "token": {
+          "type": "string"
+        },
+        "user_id": {
+          "description": "CDX user id",
+          "type": "string"
+        }
+      }
+    },
+    "SubmitTransaction": {
```
pkarman, line 406: `SignatureSubmissionResponse`?

app/services/manifest_token_json_converter.rb
pkarman, line 1: Maybe these are just private methods in the `ManifestSubmitter` service class?

pkarman, line 1: Better yet, I would mangle the params in the controller. That's really what a controller is for: translating user input into something the model can act upon.

pkarman:
@brandocalrissian some naming suggestions inline.

The `CDX::Manifest` class changes are not a part of this PR, yet it seems like that class is going to have to change (you reference that in the description). Does it make any sense to have a different CDX::\* class for this submission-via-redirect piece, and leave the current `CDX::Manifest` class as-is? Could that be part of this PR? As it is, it's not clear to me that this PR results in any actually working code, since the tests only mock request/response.

public/api-documentation/swagger.json
brandocalrissian, line 406: Sure, but there's no real signature. The names have become wonky with the API changes-- all we're doing is telling one service to submit some data to another service. We're not actually signing anything. How about "SubmissionResponse"?

brandocalrissian, line 406: If the primary noun is "submission", then it's trending to generalities. As a user, what does "submission" mean? I thought I already submitted the data to the e-Manifest system, what does it mean to create a submission again?

app/controllers/api/v0/cromerr_submit_transactions_controller.rb
brandocalrissian: We're getting a "submission to the cromerr signing service transaction" returned from the service. I'm not fond of any names for that concept, so "Submissions" is good enough if it makes more sense.

app/services/manifest_token_json_converter.rb
brandocalrissian, line 1: There was some common code essentially copied in a couple of controllers-- it didn't make sense to pull the occasionally used function up to a superclass one-controller-base-class-to-rule-them-all, so it was extracted. It doesn't need to be in the "services" directory and can move to the controllers directory.

public/api-documentation/swagger.json
brandocalrissian: Once again, the awkwardness of the purpose of the service rears its ugly head. "cromerr submit" modifies "transaction" as that's what we're creating in our post. It's a _big_, painful stretch. cromerr_submission_transaction?

Ugh.

app/services/manifest_submitter.rb
brandocalrissian: I have great difficulty taking a controller and making it load data and apply some business logic to it. The responsibilities get muddled and my controller must understand all these concepts to do its multi-pronged job.

Then again, I like to reuse services, not controllers, so I give the services all the responsibility to maintain consistent business rules and leave the controllers with the sole responsibility of translating between the outside world and something valid in our business model.

So it comes down to what's acceptable? Do we make the controller do lots of things and let the service assume all of the data coming to it has been checked, or do we let the controller do its one thing and refactor whatever service it calls as needed to keep the responsibilities clear? I'm going to push hard against having big-ol'-beefy controllers (where's the Trogdor emoji?) even if all of the Pundit docs say "throw everything you need for business rule authorization into your controller!"

brandocalrissian:
@pkarman: responses inline. Will have some changes from the suggestions in another commit to the PR. The naming has been painful, so hopefully the name changes will make more sense. We can move the implementation for the CDX client class to a differently named object (thought about that, but nothing came to mind that made sense and was consistent with the naming in the system already-- ManifestSubmission???).

Yes, more tests are needed.

Thank you for your input! This "get data in the front door" PR was already getting to big for me to want to go any further on without another set of eyes.

app/services/manifest_token_json_converter.rb
pkarman, line 1: ah ok.

for methods needed in multiple controllers, you can create a `Concern` in `app/controllers/concerns/`

app/services/manifest_submitter.rb
pkarman: Fat controllers (the Sir Topham Hatt pattern) are not good, agreed.

I think it's important that we consistently apply authorization in the same place throughout the app. So far that has been in the controller layer, since that can reject some requests more quickly with appropriate HTTP-specific responses, rather than letting errors bubble up from deeper in the model layer.

The pattern with the services so far has been to let them assume no business rules (authorization) at all, but just to translate between the app and the CDX API. Middleware.

Param mangling, authn/authz checks: those are done in the controller layer elsewhere.

public/api-documentation/swagger.json
pkarman: Let's keep CROMERR out of the API. We don't mention it anywhere else. It isn't a business function specific to that one endpoint. Isn't the entire app in service to CROMERR?

public/api-documentation/swagger.json
pkarman, line 406: I suggested `Submission` because it was a word already in use elsewhere in the app to designate that particular milestone in the overall workflow.

Our app has never signed anything. It's always acted as a proxy for the CDX service, which does the actual signing. That doesn't mean we should not use the word "sign" or "signature" -- perfectly good words to describe the end result for the user.

If we are effectively replacing the existing signature API with this one, why not replace it instead of building a 2nd API endpoint alongside it? Otherwise we end up with 2 APIs for getting a Manifest signed. The User doesn't care about the historical reasons for this particular phase of the app (API) development.

Let's change the API to make "signature" use the new CDX pattern.

brandocalrissian, line 406: I had split up the "new" API from removing the "old" API as separate tasks. If it makes it easier to follow, I can make the "new" implementation take the place of the "old" signature API in this step. It certainly makes naming simpler.

app/services/manifest_token_json_converter.rb
brandocalrissian, line 1: Concerns: so I learn about something else. As a Rails novice, these concerns look like mixins with slightly different syntax. Is that the case?

pkarman, line 1: yes a concern is a mixin.

public/api-documentation/swagger.json
brandocalrissian: :+1:
CROMERR is part of it, not quite all, but it's better to have names that make sense.

brandocalrissian:
Three VCR-based tests and a front-end->service integration test are failing.

brandocalrissian:
This is actually submitting the data to the Submit CDX service with the added signing-related JSON data in the manifest.

You have to set the "CDX_SERVER" or "CDX_SUBMIT_SERVICE_URL" environment variables in dev/production for it to talk with the right CDX environment when deployed.

app/controllers/api/v0/signatures_controller.rb
```diff
@@ -19,8 +22,15 @@ def validate_signature(content)
     run_validator(SignatureValidator.new(content))
   end
 
+  def cdx_user_role_id(manifest)
```
pkarman, line 25: this seems like it belongs on the `User` model?

pkarman:
@brandocalrissian feel free to mark the existing VCR tests as `pending` or prefix with `x` to disable. Maybe leave some comments as to why.

brandocalrissian:
Merging as the new implementation is coming down the pipe. This version may get a few new changes for demonstration purposes, but the bulk of development moving forward will be in Java/Spring.

PR #195, opened 2016-03-31
Front end work demonstration page disclaimers

scottdchristian:
Demonstration page disclaimers


brandocalrissian:
The CI failure looks to be an invalid error. If we weren't going to port this implementation to a different stack, I would insist on figuring out why the CI was giving false negatives and make it stop.

PR #199, opened 2016-04-19
Registration steps 

scottdchristian:
Added quick instructions on getting a CDX account to api-doc page


brandocalrissian:
Same invalid CI error as pr #195.

I'd prefer to use CSS styling to add indents to the text instead of the &amp;nbsp; -- you should be able to embed HTML tags and modify the style either in the CSS or inline if that's what suits your fancy. I won't insist on it presently.

